PR #42, opened 2023-06-28
Created BigQuery services wrappers

prodriguezdefino:
This wrappers will simplify the source unit and integration testing, since BigQuery does not have an emulator that can be used for that purpose.

PR #43, opened 2023-06-28
Creates the source splits, its state and the enumerator state

prodriguezdefino:
This classes represents the Flink job execution status, and will be passed around directly and in serialized states.


flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplitState.java
```diff
@@ -0,0 +1,72 @@
...
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final BigQuerySourceSplitState other = (BigQuerySourceSplitState) obj;
+        if (!Objects.equals(this.split, other.split)) {
+            return false;
+        }
+        return Objects.equals(this.offset, other.offset);
```
ywskycn, line 70: This will return false if two integer objects, even with the same value, right? I guess what we want here is just to check whether the two offset values are same.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplit.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/** A {@link SourceSplit} implementation for a BigQuery Read API stream. */
+@PublicEvolving
+public class BigQuerySourceSplit implements SourceSplit, Serializable {
+
+    private final String streamName;
+    private final Integer offset;
```
ywskycn, line 30: Would `long` be better than integer here?

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplit.java
```diff
@@ -0,0 +1,85 @@
...
+        this.offset = 0;
+    }
+
+    public BigQuerySourceSplit(String streamName, Integer offset) {
+        this.streamName = streamName;
+        this.offset = offset;
+    }
+
+    @Override
+    public String splitId() {
```
ywskycn, line 43: each stream will only have one split? As splitId needs to be unique with the scope of the source.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/services/BigQueryServiceImpl.java
```diff
@@ -0,0 +1,348 @@
...
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/** Implementation of the {@link BigQueryServices} interface that wraps the actual clients. */
+@Internal
+public class BigQueryServiceImpl implements BigQueryServices {
```
ywskycn, line 64: nit: `BigQueryServiceImpl`->`BigQueryServicesImpl`

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplit.java
prodriguezdefino: That's correct, every Flink split correlates with a read stream.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplitState.java
prodriguezdefino: Not sure if I understood this comment, but `Objects.equals` methods does check on the values boxed in the Integer objects.

```
$ jshell
|  Welcome to JShell -- Version 11.0.15
|  For an introduction type: /help intro

jshell> Integer a = 10
a ==> 10

jshell> Integer b = 10
b ==> 10

jshell> import java.util.Objects;

jshell> Objects.equals(a, b)
$4 ==> true
```

JIC, adding tests to corroborate that.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/enumerator/BigQuerySourceEnumState.java
```diff
@@ -0,0 +1,130 @@
...
+            return false;
+        }
+        if (!Objects.equals(this.assignedSourceSplits, other.assignedSourceSplits)) {
+            return false;
+        }
+        return Objects.equals(this.initialized, other.initialized);
+    }
+
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
prodriguezdefino: done, this is generated by the IDE, like with the `equals` method implementations.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/services/BigQueryServiceImpl.java
prodriguezdefino: I will address this change in #42 so is less merging/rebase later on.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplit.java
ywskycn: I may misunderstand the concept here. What does a `stream` mean here? If it represents a data source, then we should have multiple splits under one stream, right?

flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/source/enumerator/BigQuerySourceEnumStateSerializerTest.java
```diff
@@ -0,0 +1,96 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void testWrongSerializerVersion() throws IOException {
+        BigQuerySourceEnumState enumState = create();
+
+        byte[] serialized = BigQuerySourceEnumStateSerializer.INSTANCE.serialize(enumState);
+
+        BigQuerySourceEnumStateSerializer.INSTANCE.deserialize(1000, serialized);
+
+        // should never reach here
+        assert (true);
```
ywskycn, line 94: Should we fail here if we don't expect the code reach here?

prodriguezdefino: lol, that was the intent, I will correct this.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/split/BigQuerySourceSplit.java
prodriguezdefino: So a Flink split is a logical partition the source data we intent to read. When implementing a Source we need to make a decision on how that logical representation of a data portion will map with the underlying system. 

In the case of BigQuery, when using the StorageRead API the data can be read in a massively parallel way natively since the API partition the underlying data in a bunch of ReadStreams, each of those will be represented from the client library perspective as a remote iterator to which we can request a bunch of encoded rows (this can be in Avro or Arrow format). 

Since BQ already has a clear concept that logically partition the data that can be read from the storage source (a table or a table partition), we can use it to directly represent a Flink Split. 

So for this implementation we can minimally say Flink Split == BQ ReadStream. 

This will terminally mean that when reading a full BQ table or partition, we will be splitting the read in N different portions of data, each of them correlating with one BQ ReadStream following the natural definition of the client library.

ywskycn: Gotcha. Thanks for the explanation.

ywskycn (approved):

davidrabinowitz (changes requested):

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/enumerator/BigQuerySourceEnumState.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public static BigQuerySourceEnumState initialState() {
+        return new BigQuerySourceEnumState(
+                new ArrayList<>(), new ArrayList<>(), new ArrayList<>(), new HashMap<>(), false);
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 3;
```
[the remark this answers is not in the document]
prodriguezdefino: There is no `hashCode` method that takes a varargs param, there is one named `hash`. I will use that one and send the change shortly.

davidrabinowitz (approved):

PR #47, opened 2023-06-29
Adds example and distribution modules

prodriguezdefino:
Also fixes a NPE when the underlying table in BQ gets expired, now returns a proper exception.

PR #48, opened 2023-06-30
Use Open Clover for code coverage analysis

prodriguezdefino:
jacoco was interfering with surefire and failsafe making the IT tests fail when instrumenting through Jacoco. Open Clover works just fine and is compatible with codecov.


flink-connector-bigquery-examples/src/main/java/org/apache/flink/examples/gcp/bigquery/BigQueryExample.java
```diff
@@ -0,0 +1,140 @@
...
+            String projectName,
+            String datasetName,
+            String tableName,
+            String recordPropertyToAggregate,
+            String rowRestriction,
+            Integer limit)
+            throws Exception {
+
+        final StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        env.enableCheckpointing(60000L);
```
jayadeep-jayaraman, line 94: Can we get this as an user provided configuration

jayehwhyehentee: Addressed the same concern in previous PR (#47).

jayadeep-jayaraman (approved):

PR #50, opened 2023-07-02
Added test to validate recovery when StorageRead client fails

prodriguezdefino:
The most common error, when reading large streams, occurs when iterating on the ServerStream. This new test should be able to demonstrate the pipeline is able to recover the execution and obtain the expected results by restarting the read from the beginning of the split or the last checkpointed offset.


flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/fakes/StorageClientFaker.java
```diff
@@ -120,18 +133,60 @@ public Job dryRunQuery(String projectId, String query) {
             };
         }
 
+        static class FaultyIterator<T> implements Iterator<T> {
+
+            private final Iterator<T> realIterator;
+            private final Double errorPercentage;
+            private final Random random = new Random();
```
vishalkarve15, line 140: I wouldn't prefer using Random. If you absolutely have to, please use a fixed seed so that it's deterministic.

An alternative suggestion is to have it fail on every N-th invocation of `next()`.

vishalkarve15 (approved):

PR #53, opened 2023-07-04
Adds a "read from query results" example

prodriguezdefino:


flink-connector-bigquery-examples/src/main/java/com/google/cloud/flink/bigquery/examples/BigQueryExample.java
```diff
@@ -47,7 +48,8 @@
  * --agg-prop {required; record property to aggregate in Flink job} <br>
  * --restriction {optional; SQL-like filter applied at the BigQuery table before reading} <br>
  * --limit {optional; maximum records to read from BigQuery table} <br>
- * --checkpoint-interval {optional; time interval between state checkpoints in milliseconds}
+ * --checkpoint-interval {optional; time interval between state checkpoints in milliseconds} <br>
```
vishalkarve15, line 51: 1. We may need to fix the documentation to highlight that either of `bq-dataset`, etc. or `query` are required.
2. Another idea is to have a different example for query.

I like option 2 more, examples should be simple and concise.

jayehwhyehentee: Great catch. I did not realize the documentation was ambiguous. I've made some improvements to the `query` example and updated the documentation.
Based on your suggestions, I've applied 1 because having two executable classes might be confusing for the users. We dont want to execute both in the same run, and defaulting to one would mean over-riding via classpath to use the other.
I hope the updated documentation and examples are clear.

PR #64, opened 2023-12-14
Prepare poms for preview release

jayehwhyehentee:


flink-connector-bigquery-examples/pom.xml
```diff
@@ -24,7 +24,7 @@ under the License.
     <parent>
         <groupId>com.google.cloud.flink</groupId>
         <artifactId>flink-connector-bigquery-parent</artifactId>
-        <version>1.1-SNAPSHOT</version>
+        <version>0.1.0-preview</version>
```
vishalkarve15, line 27: This is fine. But the way we do it in the spark connector is we use a `${revision}` maven variable and pass it during the build (nightly, release, etc.) using `-Drevision=x.y.z`
For nightlies, we use `0.0.yyyyMMdd`

vishalkarve15 (approved):
Approved with a comment

PR #73, opened 2023-12-23
Adds helper class for unbounded read test.

prashastia:
Adds utils.py, containing helper functions for dynamic addition of data during unbounded read test.
Aslo modifies parse_logs.py to use the same for argument input.

/gcbrun

PR #74, opened 2023-12-23
Adds a new abstract class table_type.py 

prashastia:
Contains records creation and insertion methods for different types of tables.
Different table require different record types. The methods are common - string generation, long generation but we call them in different combinations which are defined in these methods. (separate for each variety of table)

/gcbrun


prashastia:
Closed as the branch is merged with PR #73

PR #75, opened 2023-12-23
Adds a python script to create a partitioned table and add initial value to it.

prashastia:
We will first create a partitioned table with expiry of about 2.5 days and add some initial values to it. 
Then we begin the dataproc job and dynamically add data to it.

/gcbrun


prashastia:
Closed as partitioned table creation is not needed per test.

PR #65, opened 2023-12-20
Adds e2e tests

prashastia:
A simple e2e tests to check read of a small BigQuery Table.

PR #76, opened 2023-12-23
Adds python script for incremental partition insertion.

prashastia:
insert_dynamic_partitions.py A python script to insert partitions incrementally into the partitioned table. Used by the e2e unbounded source testing script.

/gcbrun


cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+import time
+from absl import app
+from utils import utils
+
+
+def wait():
+    print(
+        'Going to sleep, waiting for connector to read existing, Time:'
+        f' {datetime.datetime.now()}'
+    )
```
jayehwhyehentee, line 18: Similar to comments in [#65](https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/65), let's utilize a logging framework instead of print to standard output.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+    # This is the time connector takes to read the previous rows
+    time.sleep(2.5 * 60)
```
jayehwhyehentee, line 20: Shouldn't the wait be determined by frequency of new partition discovery? It does not matter when the previous rows were read, because read streams are created for the table's snapshot at moment of read session creation. So, we need to insert new entries after the previous read session was created and before the next partition discovery. We should also mention that in the comment.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+        '--now_timestamp',
+        dest='now_timestamp',
+        help='Timestamp at the time of execution of the script.',
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        '--refresh_interval',
+        dest='refresh_interval',
+        help='.',
```
jayehwhyehentee, line 35: Please add a proper description.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+    # This is the time connector takes to read the previous rows
+    time.sleep(2.5 * 60)
+
+
+def main(argv: Sequence[str]) -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--now_timestamp',
+        dest='now_timestamp',
```
jayehwhyehentee, line 27: Not a great name. Something like `execution_timestamp` makes more sense. Btw, why do we need this argument?

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+    args = parser.parse_args(argv[1:])
+
+    # Providing the values.
+    project_name = args.project_name
+    dataset_name = args.dataset_name
+    table_name = args.table_name
+    now_timestamp = args.now_timestamp
+    now_timestamp = datetime.datetime.strptime(
+        now_timestamp, '%Y-%m-%d'
+    ).astimezone(datetime.timezone.utc)
```
jayehwhyehentee, line 70: Either use different variables, or replace with:
```
    now_timestamp = datetime.datetime.strptime(
        args.now_timestamp, '%Y-%m-%d'
    ).astimezone(datetime.timezone.utc)
```

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+        '"fields": [{"name": "name", "type": "string"},{"name": "number",'
+        '"type": "long"},{"name" : "ts", "type" : {"type" :'
+        '"long","logicalType": "timestamp-micros"}}]'
+    )
+    simple_avro_schema_string = (
+        '{"namespace": "project.dataset","type": "record","name":'
+        ' "table","doc": "Avro Schema for project.dataset.table",'
+        + simple_avro_schema_fields_string
+        + '}'
+    )
```
jayehwhyehentee, line 88: Replace with:
```
simple_avro_schema_string = (
    '{"namespace": "project.dataset","type": "record","name":'
    ' "table","doc": "Avro Schema for project.dataset.table",'
    f'{simple_avro_schema_fields_string}'
    '}'
)
```

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+
+        # This is a phase of insertion.
+        for partition_number in range(number_of_partitions):
+            threads = list()
+            # Insert via concurrent threads.
+            for thread_number in range(number_of_threads):
+                avro_file_local_identifier = avro_file_local.replace(
+                    '.', '_' + str(thread_number) + '.'
+                )
+                x = threading.Thread(
```
jayehwhyehentee, line 124: Please use a proper name. `x` is not sufficient.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed the name. We need this  argument to make sure we insert the partitions for the current date. If we provide a fixed date, then they will not be read (as partitions have been marked as closed) .

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Added.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,146 @@
...
+                        'current_timestamp': now_timestamp,
+                    },
+                )
+                threads.append(x)
+                x.start()
+            for _, thread in enumerate(threads):
+                thread.join()
+
+        time_elapsed = time.time() - start_time
+        prev_partitions_offset += number_of_partitions
```
jayehwhyehentee, line 137: Why is `prev_partitions_offset` being incremented multiple times in the same iteration?

cloudbuild/python-scripts/insert_dynamic_partitions.py
jayehwhyehentee: Current date can be found during execution using python's time or date libraries. Why do we need to take that as input?

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Correct. This wait is to make sure that any collision does not happen. Such that new rows are generated only after the previous read session was created. We wait for the next read separately.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Within the same iteration we are adding rows spread amongst 1 or more partitions.
so that at a new read, we make sure that multiple partitions are being read from.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed.

cloudbuild/python-scripts/insert_dynamic_partitions.py
jayehwhyehentee: Let's take an example.

First iteration:
```
    for number_of_partitions in partitions:   # number_of_partitions is 2
        ...
        prev_partitions_offset += 1   # prev_partitions_offset is 1
        ...
        # called avro_to_bq_with_cleanup with partition_number as 1
        # called avro_to_bq_with_cleanup with partition_number as 2
        ...
        prev_partitions_offset += number_of_partitions   # prev_partitions_offset is 3
        ...
```

Second iteration:
```
    for number_of_partitions in partitions:   # number_of_partitions is 1
        ...
        prev_partitions_offset += 1   # prev_partitions_offset is 4
        ...
        # called avro_to_bq_with_cleanup with partition_number as 4
        ...
        prev_partitions_offset += number_of_partitions   # prev_partitions_offset is 5
        ...
```

Third iteration:
```
    for number_of_partitions in partitions:   # number_of_partitions is 2
        ...
        prev_partitions_offset += 1   # prev_partitions_offset is 6
        ...
        # called avro_to_bq_with_cleanup with partition_number as 6
        # called avro_to_bq_with_cleanup with partition_number as 7
        ...
        prev_partitions_offset += number_of_partitions   # prev_partitions_offset is 8
        ...
```

So, we've skipped partition offsets 3 and 5. If that is intentional, then why?

cloudbuild/python-scripts/insert_dynamic_partitions.py
jayehwhyehentee: Then let's change the comment accordingly, or remove it. The current description is misleading.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Yeah, this is correct.
So to maintain time consistency, I took time at UTC which would be 18:30 hrs.
So (18:30 + 2 = 20:30) to (18:30 + 3 = 21:30) will generate values in partitions 20hrs, 21 hrs.

So in next phase if we generate for 18:30 + 3 - 18:30 + 4, the partitions will clash. 

I think this is getting too confusing. 
I'll fix this.

prashastia: This is fixed now.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+
+    args = parser.parse_args(argv[1:])
+
+    # Providing the values.
+    project_name = args.project_name
+    dataset_name = args.dataset_name
+    table_name = args.table_name
+
+    execution_timestamp = datetime.datetime.now(tz=datetime.timezone.utc).replace(hour=0,
```
prashastia, line 63: Now generating midnight timestamp.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,141 @@
...
+    # partitions[i] * number_of_rows_per_partition are inserted per phase.
+    partitions = [2, 1, 2]
+    # Insert 10000 - 30000 rows per partition.
+    # So, in a read up to 60000 new rows are read.
+    number_of_rows_per_partition = random.randint(1, 3) * 10000
+    #  BQ rate limit is exceeded due to large number of rows.
+    number_of_threads = 2
+    number_of_rows_per_thread = int(
+        number_of_rows_per_partition / number_of_threads
+    )
```
jayehwhyehentee, line 95: Check out floor division:`number_of_rows_per_partition // number_of_threads`.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def wait():
+    logging.info(
+        'Going to sleep, waiting for connector to read existing, Time:'
+        f' {datetime.datetime.now()}'
+    )
+    # This is a buffer time to allow the read streams to be formed.
+    # Allows conflicting conditions by making sure that
+    # new (to be generated) partitions are not part of the current read stream.
```
jayehwhyehentee, line 21:
```
Buffer time to ensure that new partitions are created after previous read session and before next split discovery.
```

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,141 @@
...
+    )
+
+    avro_file_local = 'mockData.avro'
+    table_creation_utils = utils.TableCreationUtils(
+        simple_avro_schema_string,
+        number_of_rows_per_thread,
+        table_id,
+    )
+
+    # Insert in phases.
```
jayehwhyehentee, line 104: `# Insert iteratively.`
Iteration is more appropriate than phase. Please apply that to other comments in this loop.

jayehwhyehentee (approved):

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed.

cloudbuild/python-scripts/insert_dynamic_partitions.py
prashastia: Fixed.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,138 @@
...
+        f'{simple_avro_schema_fields_string}'
+        '}'
+    )
+
+    # hardcoded for e2e test.
+    # partitions[i] * number_of_rows_per_partition are inserted per phase.
+    partitions = [2, 1, 2]
+    # Insert 10000 - 30000 rows per partition.
+    # So, in a read up to 60000 new rows are read.
+    number_of_rows_per_partition = random.randint(1, 3) * 10000
```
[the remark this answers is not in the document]
prashastia: The idea behind randomization was to introduce a possibility of different number of rows being inserted at every execution.
But yeah we could hardcode a fixed value of even take it as an argument.

cloudbuild/python-scripts/insert_dynamic_partitions.py
```diff
@@ -0,0 +1,138 @@
...
+                threads.append(thread)
+                thread.start()
+            for _, thread in enumerate(threads):
+                thread.join()
+
+        time_elapsed = time.time() - start_time
+        prev_partitions_offset += number_of_partitions
+        # We wait for the refresh to happen
+        # so that the data just created can be read.
+        while time_elapsed < float(60 * 2 * refresh_interval):
```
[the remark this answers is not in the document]
prashastia: The step before this includes the generation and insertion of rows to the BQ table. The time for this insertion is variable. In order to make sure next insertion does not take place before the next read stream formation, we explicitly wait for the stipulated amount of time to pass.

vishalkarve15 (approved):

PR #77, opened 2023-12-23
Adds shell script for unbounded table read , modifies table_read to execute the same.

prashastia:
Adds shell script that

Creates the asynchronously running unbounded job - the dataproc job is created in mode unbounded
Dynamically adds partitions - insert_dynamic_partitions.py is executed with the necessary parameters.
kills the dataproc job - As the read has completed and its correctness needs to be checked further.

/gcbrun


cloudbuild/e2e-test-scripts/table_read.sh
```diff
@@ -36,7 +36,9 @@ if [ "$MODE" == "bounded" ]
 then
   echo "Bounded Mode!"
   source cloudbuild/e2e-test-scripts/bounded_table_read.sh
-
+else
+  echo "Unbounded Mode!"
+  source cloudbuild/e2e-test-scripts/unbounded_table_read.sh
```
jayehwhyehentee, line 41: We should explicitly check if mode is `unbounded`, just like the `bounded` case. If mode is neither bounded nor unbounded, throw error.

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Running this job async to make sure it exits so that dynamic data can be added
+gcloud dataproc jobs submit flink --id "$JOB_ID" --jar="$GCS_JAR_LOCATION" --cluster="$CLUSTER_NAME" --region="$REGION" --async -- --gcp-project "$PROJECT_NAME" --bq-dataset "$DATASET_NAME" --bq-table "$TABLE_NAME" --mode unbounded --agg-prop "$AGG_PROP_NAME" --ts-prop "$TS_PROP_NAME" --partition-discovery-interval "$PARTITION_DISCOVERY_INTERVAL"
+
+# Dynamically adding the data. This is timed 2.5 min wait for read and 5 min refresh time.
```
jayehwhyehentee, line 20: Were you able to find the right combination of wait and refresh intervals?

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
```diff
@@ -0,0 +1,28 @@
...
+python3 cloudbuild/python-scripts/insert_dynamic_partitions.py -- --project_name "$PROJECT_NAME" --dataset_name "$DATASET_NAME" --table_name "$TABLE_NAME" --refresh_interval "$PARTITION_DISCOVERY_INTERVAL"
+
+# Wait for a bit, as mapping and output of records takes some time.
+sleep 3m
```
jayehwhyehentee, line 24: How did we arrive at this number?

prashastia: In our code we wait 2.5 mins for the read streams to form (prior to insertion). 
This also makes sure that the previously inserted records are read properly.
But after the last insertion, since there is no wait in the python code, we wait through the script.

Can remove the extra 30 seconds.

cloudbuild/e2e-test-scripts/table_read.sh
```diff
@@ -36,7 +36,13 @@ if [ "$MODE" == "bounded" ]
 then
   echo "Bounded Mode!"
   source cloudbuild/e2e-test-scripts/bounded_table_read.sh
-
+elif [ "$MODE" == "unbounded" ]
```
jayehwhyehentee, line 39: Let's use shell `case` statement, as [here](https://github.com/GoogleCloudDataproc/flink-bigquery-connector/blob/main/cloudbuild/presubmit.sh#L30)

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
jayehwhyehentee: Having a cushion is alright. Just wanted to understand the reason behind this number.

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
jayehwhyehentee: Satisfied with offline discussion. Resolving this.

jayehwhyehentee (approved):

PR #81, opened 2023-12-26
Adds python script to list all running jobs in a cluster and cancel them.

prashastia:
1. Lists all the running job in a cluster in a region.
2. Then cancel all such running jobs.

/gcbrun


prashastia:
Closed as jobs no longer need cancellation.

PR #80, opened 2023-12-25
Adds script to create the clusters for the nightly e2e jobs.

prashastia:
Adds script to create the clusters for the nightly e2e jobs.
1. gcloud command to create the temp, staging buckets and creating the cluster.
2. Adds the step in `nightly.yaml` for the cluster creation.
3. Call the above-mentioned script from `nightly.sh`.

/gcbrun

PR #82, opened 2023-12-26
Adds shell script that cancels running jobs, deletes the cluster and its temp and storage buckets.

prashastia:
1. Calls the python script `cancel_dataproc_jobs.py` that cancels all the running jobs in a cluster and region.
2. Deletes the cluster
3. Also deletes its staging and temp buckets.

/gcbrun


prashastia:
Closed as this functionality is already supported by PR #80

PR #83, opened 2023-12-26
Adds shell script that executes job cancellation, cluster and bucket deletion.

prashastia:
1. Adds `nightly_cancel.sh` - it executes the script `cancel_job_delete_cluster.sh` for all of the clusters.

/gcbrun


prashastia:
Closed as no longer needed.

PR #88, opened 2024-01-18
Fixes the unbounded read test for the nightly job.

prashastia:
- The current solution is the use a KeyedProcessFunction and use the `processElement()` method to count the number of records that have been processed. 
- Whenever the expected count is reached, the method logs. Count exceed is also logged. 
- The job is terminated after a timeout of 18 minutes. 
- Since KeyedProcessFunction is used for the counting, same key is used i.e a key named "commonKey" is added to the records to make sure they are received together. 
- To solve the problem of parallel operator, parallelism and number of task slots have been set as 1.
---
To check for the job success:
- We check if there are any logs pertaining to exceeded count. In such a case a failure is reported. 
- In case of absence of "count-exceeded" logs, we search for the actual count read. We then check this count with the number of rows in the BQ table. In case of a match, the job succeeds.
- In case a less number of records are read then we raise an exception indicating that the logs are actually not present and less number of records being processed can be a possible cause. 

/gcbrun


cloudbuild/e2e-test-scripts/bounded_table_read.sh
```diff
@@ -14,7 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-# We won't run this async as we can wait for a bounded job to succeed or fail.
-gcloud dataproc jobs submit flink --id "$JOB_ID" --jar="$GCS_JAR_LOCATION" --cluster="$CLUSTER_NAME" --region="$REGION" -- --gcp-project "$PROJECT_NAME" --bq-dataset "$DATASET_NAME" --bq-table "$TABLE_NAME" --agg-prop "$AGG_PROP_NAME" --query "$QUERY_STRING"
+# Adds retries in case same job-id has been randomly generated again.
```
prashastia, line 17: This can be removed once the cluster creation script has been merged.
But prior to merging, can be present as a precaution.

jayehwhyehentee:
```
1. We check if the records are actually absent by retrying to search for the logs.
2. In case after 5 retries the records are still absent, then we raise an exception indicating that the logs are actually not present.
```

We don't need this if the wait has been increased to 1 minute. I believe the logs are found after a minute, and retries are not required. Please correct me if I'm wrong.

prashastia:
> ```
> 1. We check if the records are actually absent by retrying to search for the logs.
> 2. In case after 5 retries the records are still absent, then we raise an exception indicating that the logs are actually not present.
> ```
> 
> We don't need this if the wait has been increased to 1 minute. I believe the logs are found after a minute, and retries are not required. Please correct me if I'm wrong.

Yes, this is fixed now.

cloudbuild/e2e-test-scripts/bounded_table_read.sh
jayehwhyehentee: I don't think we need to worry about this given the short period of time till the per test cluster feature is merged. Please remove these retries.

cloudbuild/e2e-test-scripts/table_read.sh
```diff
@@ -40,6 +40,8 @@ elif [ "$MODE" == "unbounded" ]
 then
   echo "Unbounded Mode!"
   source cloudbuild/e2e-test-scripts/unbounded_table_read.sh
+  # Add more sleep time. as IO might take time.
+  sleep 55
```
jayehwhyehentee, line 44: Why 55 seconds? Why not a round number like 1 minute?

jayehwhyehentee: Because you plan to sleep 5 seconds at the end?

jayehwhyehentee: Its still better to use 60 seconds here. Prefer round numbers for arbitrary usage.

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
```diff
@@ -14,15 +14,24 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Get the timestamp to annotate the copy of the partitioned table.
+timestamp=$(date +"%Y%m%d%H%M%S")
+
+# Copy the table
+bq cp -f "$DATASET_NAME"."$TABLE_NAME" "$DATASET_NAME"."$TABLE_NAME"_"$timestamp"
+
+# Change the table name for all future uses.
```
jayehwhyehentee, line 23: `# Rename the table copy name for using in this test.`

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
```diff
@@ -14,15 +14,24 @@
...
+TABLE_NAME="$TABLE_NAME"_"$timestamp"
+
+# Set the expiration time to 1 hour.
+bq update --expiration 3600 "$DATASET_NAME"."$TABLE_NAME"
```
jayehwhyehentee, line 27: Good job!

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
```diff
@@ -14,15 +14,24 @@
...
 
-# Wait for a bit, as mapping and output of records takes some time.
-sleep 3m
-
-# Kill the dataproc job
-# quiet flag will not ask for console input.
-gcloud dataproc jobs kill "$JOB_ID" --region="$REGION" --quiet
+# Now the Dataproc job will automatically succeed after stipulated time (18 minutes hardcoded).
+# we wait for it to succeed or finish.
+gcloud dataproc jobs wait "$JOB_ID" --region "$REGION" --project "$PROJECT_NAME"
```
jayehwhyehentee, line 37: Looking at https://cloud.google.com/sdk/gcloud/reference/dataproc/jobs/wait, seems like it gets you the output of the job. It doesn't mention anything about terminating it. Please follow up with someone from Dataproc to verify your usage is correct and result will always be as intended.

cloudbuild/python-scripts/parse_logs.py
```diff
@@ -190,6 +194,41 @@ def get_blob_and_check_metric(
...
 
+    if mode == 'unbounded':
+        '''
+        Check if blob containing the logs contain the string of the form 
+        "Number of records processed (%d) exceed the expected count (%d)"
+        then throw an error,
+        If not get logs of type  "%d number of records have been processed".
+        Read the blob and get the metric from them,
+        and check if they match the bigquery table count.     
+        '''
```
jayehwhyehentee, line 205: This is same as whats mentioned in comments below. Please remove this.

cloudbuild/python-scripts/parse_logs.py
```diff
@@ -190,6 +194,41 @@ def get_blob_and_check_metric(
...
+        # This is hardcoded on the basis of logs
+        records_exceed_log_pattern = (r'Number of records processed \((\d+)\) exceed the expected '
+                                      r'count \((\d+)\)')
+        match = re.search(records_exceed_log_pattern, logs_as_string)
+
+        # Check if blob containing the logs contain the string of the form
+        # "Number of records processed (%d) exceed the expected count (%d)"
+        # then throw an error.
```
jayehwhyehentee, line 213: `If logs contain the string mentioned below, then count of records has exceeded the expected count, so throw an error.`

cloudbuild/e2e-test-scripts/unbounded_table_read.sh
prashastia: Correct. 
We just wait for the job to finish on its own.
Since the java application has now been modified to cancel on its own after timeout of 18 minutes.

jayehwhyehentee: Ah makes sense. Thanks for clarifying.

cloudbuild/python-scripts/parse_logs.py
```diff
@@ -324,34 +354,44 @@ def read_logs(cluster_temp_bucket, logs_pattern, query):
...
         # -1 is returned in case metric not found in the log file.
-        (metric_value, is_query_result_present) = get_blob_and_check_metric(
+        (metric_value, is_query_result_present,
+         is_unbounded_result_present) = get_blob_and_check_metric(
             gcs_log_object,
             cluster_temp_bucket,
             metric_string,
             end_of_metric_string,
-            query,
+            query, mode
```
jayehwhyehentee, line 382: Lets be consistent and add `mode` on new line

cloudbuild/python-scripts/parse_logs.py
```diff
@@ -324,34 +354,44 @@ def read_logs(cluster_temp_bucket, logs_pattern, query):
...
         if metric_value != -1:
             is_metric_found = True
             # Sum up all the values.
             total_metric_count += metric_value
 
+        # Check if unbounded read result is found in any one of the files.
+        # True if found in at least one of the files.
+        # False in case not found in any.
+        if is_unbounded_result_present:
+            is_unbounded_result_found = True
```
jayehwhyehentee, line 393: These flags seem redundant. Just `is_unbounded_result_present` can be used without introducing `is_unbounded_result_found `.

flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -184,13 +197,74 @@ private static void runJob(
...
+                                    throws Exception {
+
+                                this.numRecords.update(this.numRecords.value() + 1);
+
+                                if (this.numRecords.value() > expectedValue) {
+                                    LOG.info(
+                                            String.format(
+                                                    "Number of records processed (%d) exceed the expected count (%d)",
+                                                    this.numRecords.value(), expectedValue));
+
```
jayehwhyehentee, line 239: Remove extra line.

flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -184,13 +197,74 @@ private static void runJob(
...
+                            public void processElement(
+                                    Tuple2<String, Integer> value,
+                                    KeyedProcessFunction<String, Tuple2<String, Integer>, Long>
+                                                    .Context
+                                            ctx,
+                                    Collector<Long> out)
+                                    throws Exception {
+
+                                this.numRecords.update(this.numRecords.value() + 1);
+
```
jayehwhyehentee, line 233: Please remove extra lines.

flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -184,13 +197,74 @@ private static void runJob(
...
+                                                "numRecords", TypeInformation.of(Long.class), 0L);
+                                this.numRecords = getRuntimeContext().getState(descriptor);
+                            }
+
+                            @Override
+                            public void processElement(
+                                    Tuple2<String, Integer> value,
+                                    KeyedProcessFunction<String, Tuple2<String, Integer>, Long>
+                                                    .Context
+                                            ctx,
```
jayehwhyehentee, line 228: This indentation does not look right.

cloudbuild/python-scripts/parse_logs.py
prashastia: `is_unbounded_result_present` is the local variable that gets the result from one log file. `is_unbounded_result_found ` is an OR across all the log files as we need the result to be present in atleast one log file.

jayehwhyehentee: Got it

jayehwhyehentee (approved):

PR #95, opened 2024-03-03
Adds Serialisation Methods: Obtain Descriptor from Avro Schema.

prashastia:
Adds Serialisation capabilities to convert PRIMITIVE Avro Types(INT, BOOL, DOUBLE, ..) to Storage Api Protos.
The logic is to obtain the table Schema from the input BigQuery table, convert it into avro schema and then obtain the Descriptor for it.


prashastia:
Obsolete.

PR #100, opened 2024-03-12
Implements conversion of `MAP` and `ARRAY` avro type to be converted to Bigquery Descriptors.

prashastia:
* MAP Conversion:  A`MAP` has been interpreted as `RECORD` field with `MODE` as `REPEATED` in BigQuery.
    * `ARRAY` of type `MAP` is not supported (Record with 2 fields **KEY** and **VALUE**).
    * `MAP` in `UNION` type is not supported yet.
* ARRAY Conversion: An `ARRAY` is made with the provided datatype with `MODE` as `REPEATED` in BigQuery.
    * `ARRAY` of Type `UNION` is not supported.
    *  Nested arrays are not supported.
    * To summarize: `OPTIONAL` and `REPEATED` type fields are not supported in an array.
   
Note: The current implementation also would not be able to handle nesting of `RECORD` datatype > 15 levels (bigQuery maximum permitted). In case of > 15 layers, the Descriptor would be formed, but error would be thrown during insertion.

/gcbrun


prashastia:
obsolete

PR #104, opened 2024-04-01
Adds support for `serialize()` for Primitive, `RECORD`, `UNION` and `ARRAY` datatypes.

prashastia:
The implementation allows actual avro records to be serialized into storage api proto transferable protobufs.

The mapping is as follows:
1. Primitive Types - Object value is checked against the schema type passed and converted to its respective type. e.g If the schema indicates a `STRING` field, the value is validated to be a string and return the passed object as `STRING`.
2.  MAP - Not supported yet.
3. UNION - Validation is done to ensure the absence of multiple not null or only null type. In case of type of format ["null", datatype] the datatype value is converted. `null` values are allowed in this case.
4. ARRAY - Individual values are validated and converted one by one. In case of `null` value in array,  Error is thrown.
5. RECORD -  Recursive approach to convert each field items one by one. A `MESSAGE` is returned containing the respective fields.

/gcbrun

PR #105, opened 2024-04-10
Default sink implementation

jayehwhyehentee:
Create Sink and Writer implementations for using BigQuery's default stream. This mode is intended to offer at-least-once consistency guarantee.

/gcbrun


flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryBaseSink.java
```diff
@@ -18,5 +18,58 @@
...
+        if (sinkConfig.getSerializer() == null) {
+            throw new IllegalArgumentException("BigQuery serializer cannot be null");
+        }
+        if (sinkConfig.getSchemaProvider() == null) {
+            throw new IllegalArgumentException("BigQuery schema provider cannot be null");
+        }
+    }
+
+    /** Ensures Sink's parallelism does not exceed the allowed maximum when scaling Flink job. */
+    void checkParallelism(int numberOfParallelSubtasks) {
```
clmccart, line 66: shouldn't we check parallelism as part of validating the sink config? no need to wait until we try to create writers

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQuerySink.java
```diff
@@ -0,0 +1,38 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Class wrapping BigQuery sinks with appropriate configurations. */
+public class BigQuerySink {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BigQuerySink.class);
+
+    public static Sink get(BigQuerySinkConfig sinkConfig, StreamExecutionEnvironment env) {
+        if (sinkConfig.getDeliveryGuarantee() == DeliveryGuarantee.EXACTLY_ONCE) {
```
clmccart, line 32: what if delivery guarantee is NONE? https://nightlies.apache.org/flink/flink-docs-master/api/java/org/apache/flink/connector/base/DeliveryGuarantee.html#NONE

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/exceptions/BigQueryConnectorException.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.flink.bigquery.sink.exceptions;
+
+/** This class wraps errors found while connecting to BigQuery storage. */
+public class BigQueryConnectorException extends RuntimeException {
```
clmccart, line 20: i'm sure there's a reason but why wrap these exceptions in a new type if we arent enriching them?

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/AvroToProtoSerializer.java
```diff
@@ -27,4 +27,9 @@ public class AvroToProtoSerializer implements BigQueryProtoSerializer<GenericRec
     public ByteString serialize(GenericRecord record) throws BigQuerySerializationException {
         throw new UnsupportedOperationException("serialize method is not supported");
     }
+
+    @Override
+    public void init(BigQuerySchemaProvider schemaProvider) {
+        throw new UnsupportedOperationException("init method is not supported");
```
clmccart, line 33: do callers not need to invoke init? ie, is the constructor sufficient? if so, we should include that in the error message.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
```diff
@@ -0,0 +1,205 @@
...
+import com.google.cloud.flink.bigquery.sink.exceptions.BigQuerySerializationException;
+import com.google.cloud.flink.bigquery.sink.serializer.BigQueryProtoSerializer;
+import com.google.cloud.flink.bigquery.sink.serializer.BigQuerySchemaProvider;
+import com.google.protobuf.ByteString;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.LinkedList;
+import java.util.Queue;
```
clmccart, line 41: do we need the queue and the linkedlist?

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * Base class for developing a BigQuery writer.
+ *
+ * <p>This class abstracts implementation details common for BigQuery writers which use the {@link
+ * StreamWriter}.
+ *
```
clmccart, line 48: can we add to this doc comment a general outline of how bytes are expected to flow through this writer?

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BigQueryDefaultWriter.java
```diff
@@ -40,20 +47,67 @@
  *
```
clmccart, line 47: same comment as above. for example, when write is called, data gets added to the buffer and is asynchronously appended to BQ row when flushed, etc.

flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/sink/serializer/TestBigQuerySerializer.java
```diff
@@ -0,0 +1,54 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.flink.bigquery.sink.serializer;
+
+import com.google.cloud.flink.bigquery.sink.exceptions.BigQuerySerializationException;
+import com.google.protobuf.ByteString;
+
+/** Mock serializer for Sink unit tests. */
```
clmccart, line 22: Let's call this a "Fake" to be consistent

clmccart (changes requested):

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryBaseSink.java
jayehwhyehentee: We've moved the parallelism out of sink config because:
1. it is easy to bypass for the user.
2. scaling up the cluster would require another parallelism check which can only be done when creating new writers.
3. first time writer creation happens during job initialization in the beginning so it would be done at the same time as validating sink config.

Writers are created without write streams so we're not invoking BQ storage APIs at all during initialization.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQuerySink.java
jayehwhyehentee: The default sink will offer either at-least-once or at-most-once depending on the checkpointing mode. Eventual data consistency in the destination (like BQ table) is governed by the combination of sink delivery guarantee and checkpointing mode.

1. If checkpointing is disabled, then regardless of delivery guarantee the sink will offer at-most-once consistency.
2. For default sink to offer at-least-once, delivery guarantee must be at-least-once and checkpointing mode should be at-least-once or exactly-once.
3. Delivery guarantee must be exactly-once isnt supported yet.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/exceptions/BigQueryConnectorException.java
jayehwhyehentee: These wrapper exceptions will be enriched (similar to [spark conector](https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/master/bigquery-connector-common/src/main/java/com/google/cloud/bigquery/connector/common/BigQueryConnectorException.java)) when more complex execution states are introduced with exactly-once.

Same reasoning applies to the serialization exception. That class will be enriched in upcoming iterations, post MVP.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/AvroToProtoSerializer.java
jayehwhyehentee: Implementation for `init` has been added in #104 ([here](https://github.com/GoogleCloudDataproc/flink-bigquery-connector/blob/b0ae6c26143991f2b7ac2c747d42e11f7599950e/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/AvroToProtoSerializer.java#L54)).

`do callers not need to invoke init?`
We dont want to enforce that. The `init` is an opinionated action that our AvroToProtoSerializer needs to perform based on the BigQuery table's schema. Other serializers (which can be defined by users) may not need this "pre-process" and hence it can be a no-op. Also, we do not want to tie the `BigQuerySchemaProvider` with the serializer's constructor for similar reasons. The schema provider is used for multiple purposes by the sink and we are trying to hide most of it from the user. Hence, the serializer and schema provider are separate members in the sink config.

`is the constructor sufficient?`
If the respective serializer's design is sufficient with just the constructor and no init, then we want to allow that, basically making the `init` a no-op.

jayehwhyehentee: That being said, would you suggest making the parent `BigQueryProtoSerializer` interface an abstract class instead where the `init` method is empty instead of abstract?

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
jayehwhyehentee: Queue is an interface and LinkedList seems to be the most standard implementation. We're storing the append response futures in a queue to enforce that only `add`, `peek` and `poll` are used in the writer implementation.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/AvroToProtoSerializer.java
jayehwhyehentee: Made the above change of defining `BigQueryProtoSerializer` as abstract class instead of interface and making `init` method a no-op unless over-ridden.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
jayehwhyehentee: Done. Lmk if you'd like me to add more details here.

flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/sink/BigQuerySinkTest.java
```diff
@@ -0,0 +1,56 @@
...
+                BigQuerySinkConfig.newBuilder()
+                        .connectOptions(StorageClientFaker.createConnectOptionsForWrite(null))
+                        .schemaProvider(TestBigQuerySchemas.getSimpleRecordSchema())
+                        .serializer(new TestBigQuerySerializer(ByteString.copyFromUtf8("foo")))
+                        .build();
+        assertNotNull(BigQuerySink.get(sinkConfig, null));
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void testExactlyOnceNotSupported() throws IOException {
```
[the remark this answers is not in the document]
jayehwhyehentee: Done.

flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/sink/serializer/TestBigQuerySerializer.java
jayehwhyehentee: Done. Good catch, thanks!

jayehwhyehentee:
@clmccart Details regarding sink's behavior wrt delivery guarantee and checkpointing will be mentioned in the root readme. Concise descriptions have been added as javadocs in BigQueryDefaultSink and BigQuerySink.

jayehwhyehentee:
@agrawal-siddharth would love to get your thoughts on this PR.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryBaseSink.java
clmccart: okay that's fine. you can resolve this convo; thanks!

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQuerySink.java
clmccart: yeah that makes sense to me but the user can pass a deliveryGuarentee.NONE during configuration. i'm wodnering if we want to throw an error there or if we want to just treat it as ALO

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/AvroToProtoSerializer.java
clmccart: nice i like that better; thanks! you can resolve this convo

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
clmccart: okay that's fine. you can resolve this convo; thanks!

flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/sink/BigQuerySinkTest.java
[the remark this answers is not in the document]
clmccart: you can resolve this convo; thanks!

flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/sink/serializer/TestBigQuerySerializer.java
clmccart: you can resolve this convo; thanks!

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BigQueryDefaultWriter.java
clmccart: i think what oyu have is sufficient (in conjuncture w/ the javadoc on the basewriter). you can resolve this!

clmccart (approved):

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BigQueryDefaultWriter.java
```diff
@@ -40,20 +47,67 @@
...
+        }
+        ApiFuture responseFuture = streamWriter.append(protoRowsBuilder.build());
+        appendResponseFuturesQueue.add(responseFuture);
     }
 
+    /** Throws a RuntimeException if an error is found with append response. */
     @Override
-    public void close() {
-        throw new UnsupportedOperationException("close method is not supported");
+    void validateAppendResponse(ApiFuture<AppendRowsResponse> appendResponseFuture) {
```
agrawal-siddharth, line 89: Do you anticipate any difference in this method for the exactly-once writer? If not, then this could go on the base class?

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQuerySink.java
jayehwhyehentee: Good point. Let's restrict to at-least-once for now, and we can extend this sink to support deliveryGuarentee.NONE in future.
Just to summarize though, theoretically this same sink can offer both at-least-once and at-most-once based on checkpointing mode.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BigQueryDefaultWriter.java
jayehwhyehentee: Yes there will be some differences for the exactly-once writer, such as handling of offset related errors.

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryBaseSink.java
```diff
@@ -18,5 +18,57 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /** Base class for developing a BigQuery sink. */
-abstract class BigQueryBaseSink implements Sink {}
+abstract class BigQueryBaseSink implements Sink {
+
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public static final int MAX_SINK_PARALLELISM = 100;
```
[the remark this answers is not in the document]
jayehwhyehentee: Done

prashastia (approved):

flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BigQueryDefaultWriter.java
```diff
@@ -40,20 +47,67 @@
...
+    /** Asynchronously append to BigQuery table's default stream. */
     @Override
-    public void flush(boolean endOfInput) throws IOException, InterruptedException {
-        throw new UnsupportedOperationException("flush method is not supported");
+    void sendAppendRequest() {
+        if (streamWriter == null) {
+            streamWriter = createStreamWriter(true);
+        }
+        ApiFuture responseFuture = streamWriter.append(protoRowsBuilder.build());
+        appendResponseFuturesQueue.add(responseFuture);
```
agrawal-siddharth, line 84: It feels odd that the derived class is manipulating state owned by the base class. How about modifying sendAppendRequest to have the following signature:

ApiFuture sendAppendRequest(ProtoRows protoRows)

Then the implementation is responsible for sending the given rows and returning the future. The base class would then take care of generating the rows, and updating the appendResponseFuturesQueue.

jayehwhyehentee: Thats a great suggestion! Applied to the implementation. Thank you

agrawal-siddharth (approved):

PR #114, opened 2024-04-30
Add distribution management in pom for nexus

jayehwhyehentee:
Inferred from spark bq connector: https://github.com/GoogleCloudDataproc/spark-bigquery-connector/blob/master/spark-bigquery-parent/pom.xml#L39


prodriguezdefino (approved):
LGTM

PR #137, opened 2024-07-01
Implementation for `BigQueryDynamicSink` and `BigQueryDynamicSinkFactory`

prashastia:
/gcbrun


flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactory.java
```diff
@@ -102,11 +113,35 @@ public DynamicTableSource createDynamicTableSource(Context context) {
             config = config.withTestingServices(testingServices);
         }
 
+        // Create an unbounded source.
```
jayehwhyehentee, line 116: Combine into one method, and use ternary operator to distinguish between the two modes.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactory.java
```diff
@@ -102,11 +113,35 @@ public DynamicTableSource createDynamicTableSource(Context context) {
...
+        if (config.isUnboundedEnabled()) {
+            return new BigQueryDynamicTableSource(
+                    config.toBigQueryReadOptions(),
+                    context.getPhysicalRowDataType(),
+                    Boundedness.CONTINUOUS_UNBOUNDED);
+        }
+
         return new BigQueryDynamicTableSource(
                 config.toBigQueryReadOptions(), context.getPhysicalRowDataType());
     }
```
jayehwhyehentee, line 126: Combine into one method, and use ternary operator to distinguish between the two modes.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
```diff
@@ -0,0 +1,124 @@
...
+                && (this.sinkConfig.getConnectOptions() == object.sinkConfig.getConnectOptions())
+                && (this.sinkConfig.getSerializer() == object.sinkConfig.getSerializer())
+                && (this.sinkConfig.getDeliveryGuarantee()
+                        == object.sinkConfig.getDeliveryGuarantee())) {
+            BigQuerySchemaProvider thisSchemaProvider = this.sinkConfig.getSchemaProvider();
+            BigQuerySchemaProvider objSchemaProvider = object.sinkConfig.getSchemaProvider();
+            return thisSchemaProvider.getAvroSchema().equals(objSchemaProvider.getAvroSchema());
+        }
+        return false;
+    }
```
jayehwhyehentee, line 89: Lets not override such base methods just for testing. Define your own equality checking method in the test file.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public ChangelogMode getChangelogMode(ChangelogMode requestedMode) {
+        return ChangelogMode.insertOnly();
+    }
```
jayehwhyehentee, line 94: Why are we hardcoding this? Where is `requestedMode` coming from? If it is user provided, then `non insertOnly` modes should be returned with unsupported operation or illegal argument exception

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactoryTest.java
```diff
@@ -95,6 +103,42 @@ public void testBigQueryReadProperties() throws IOException {
...
+    @Test
+    public void testBigQueryUnboundedReadProperties() throws IOException {
+        Map<String, String> properties = getRequiredOptions();
+        properties.put(BigQueryConnectorOptions.COLUMNS_PROJECTION.key(), "aaa,bbb");
+        properties.put(BigQueryConnectorOptions.MAX_STREAM_COUNT.key(), "100");
+        properties.put(
+                BigQueryConnectorOptions.ROW_RESTRICTION.key(), "aaa > 10 AND NOT bbb IS NULL");
+        properties.put(
+                BigQueryConnectorOptions.SNAPSHOT_TIMESTAMP.key(),
+                "" + Instant.EPOCH.toEpochMilli());
```
jayehwhyehentee, line 115: Use `Long.toString()` for converting long to string

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkITCase.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Google Inc.
+ * Copyright (C) 2024 Google Inc.
```
jayehwhyehentee, line 2: No need to change this

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactory.java
prashastia: Fixed.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
prashastia: Fixed.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactoryTest.java
prashastia: Fixed.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkITCase.java
prashastia: ACK.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
prashastia: [Source](https://github.com/apache/flink/blob/48f352208b5cb3c15202a12b2f8ba5f6f9eab6a1/flink-table/flink-table-common/src/main/java/org/apache/flink/table/connector/sink/DynamicTableSink.java#L83) The sink must specify the kind of changelog(`DELETE`, `INSERT`, `UPDATE_AFTER`, `UPDATE_BEFORE`) it supports. It is the planner that requests a expected mode according to plan and in case it is not supported, error is thrown.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public SinkRuntimeProvider getSinkRuntimeProvider(Context context) {
+        // init() should be called itself.
+        // Set the logical type.
+        ((RowDataToProtoSerializer) sinkConfig.getSerializer()).setLogicalType(this.logicalType);
+        // Get the Datastream-API Sink.
+        return SinkV2Provider.of(BigQuerySink.get(this.sinkConfig, null));
```
jayehwhyehentee, line 64: The second argument to `BigQuerySink.get` should not be null. Is it possible to get the `StreamExecutionEnvironment` here? Store it in some context?

yerramalli: @jayehwhyehentee  BigQuerySink.get is not using the StreamExecutionEnvironment anywhere,  It is not even stored in a member variable or used in the constructor. Why is it needed?

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public DynamicTableSink copy() {
+        return new BigQueryDynamicTableSink(this.sinkConfig, this.logicalType);
+    }
+
+    @Override
+    public String asSummaryString() {
+        return "BigQuery";
```
yerramalli, line 74: At some point we should consider moving these hardcoded strings into a resource bundle file.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/fakes/StorageClientFaker.java
```diff
@@ -580,57 +580,70 @@ public static BigQueryReadOptions createReadOptions(
...
                 avroSchemaString,
-                params -> StorageClientFaker.createRecordList(params),
+                dataGenerator,
+                appendRowsResponse,
                 0D,
                 -1);
     }
 
-    public static BigQueryReadOptions createTableReadOptions(
+    public static BigQueryReadOptions createReadAndWriteOptions(
```
yerramalli, line 600: What is the reason for changing the method to create a read and write options? Btw, it is still returning a BiqQueryReadOptions object. Something does not seem right here.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkITCase.java
```diff
@@ -45,10 +49,11 @@
...
 import java.time.Instant;
 import java.util.List;
-import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static com.google.cloud.flink.bigquery.fakes.StorageClientFaker.createConnectOptionsForWrite;
+
 /** An integration test for the SQL interface of the BigQuery connector. */
 public class BigQueryDynamicTableSinkITCase {
```
yerramalli, line 58: Please use verbose names, this is a integration test so have that in the name. IT is very cryptic.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Method overwritten to check equality, required for testing.
+     *
+     * @param bigQueryDynamicSink1 Object 1.
+     * @param bigQueryDynamicSink2 Target Object to check equality with.
+     * @return True if {@link Object} is equal to current object.
+     */
+    private boolean checkEquality(Object bigQueryDynamicSink1, Object bigQueryDynamicSink2) {
```
yerramalli, line 132: This is not a override of the equals method, it is test specific, right? If this is needed in other places also, I would suggest you consider overriding the equals method on the BigQueryDynamicTableSink class.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
```diff
@@ -0,0 +1,157 @@
...
+        if (bigQueryDynamicSink1 == bigQueryDynamicSink2) {
+            return true;
+        }
+        if (bigQueryDynamicSink1.getClass() != bigQueryDynamicSink2.getClass()) {
```
yerramalli, line 136: This code is partly broken as bigQueryDynamicSink1 and bigQueryDynamicSink2 can be instances of the same class but not necessarily be a BigQueryDynamicTableSink instance. If the passed in arguments are not instances of BigQueryDynamicTableSink then the subsequent code where you are directly doing a type cast would throw a ClassCastException.

Instead of checking the class equality use an instanceof check with BigQueryDynamicTableSink on both arguments.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
```diff
@@ -0,0 +1,157 @@
...
+            return false;
+        }
+        BigQueryDynamicTableSink bigQueryDynamicTableSink1 =
```
yerramalli, line 139: Looking at this implementation it looks like the equality is mainly on the config and not the actual sink object. Why not override the equals method on the config class itself?

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
```diff
@@ -0,0 +1,157 @@
...
+                (BigQueryDynamicTableSink) bigQueryDynamicSink2;
+        BigQuerySinkConfig sinkConfig1 = bigQueryDynamicTableSink1.getSinkConfig();
+        BigQuerySinkConfig sinkConfig2 = bigQueryDynamicTableSink2.getSinkConfig();
+
+        if ((bigQueryDynamicTableSink1.getLogicalType()
+                        == bigQueryDynamicTableSink2.getLogicalType())
+                && (sinkConfig1.getConnectOptions() == sinkConfig2.getConnectOptions())
+                && (sinkConfig1.getSerializer() == sinkConfig2.getSerializer())
+                && (sinkConfig1.getDeliveryGuarantee() == sinkConfig2.getDeliveryGuarantee())) {
+            BigQuerySchemaProvider thisSchemaProvider = sinkConfig1.getSchemaProvider();
```
yerramalli, line 151: Be consistent with the variable names. Instead of 'thisSchemaProvider' just use 'schemaProvider1'. As this is not an instance method, no need to use names that start with 'this'.

yerramalli (changes requested):

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
jayehwhyehentee: It will be used when we implement exactly-once. This is the only method customers are supposed to invoke, so we defined the signature such that future releases will not break customer usage.
Also, there are some validations which we plan to add for at-least-once in the future which will read StreamExecutionEnvironment, and again we don't want to break signature.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
jayehwhyehentee: That was the original implementation, however I disagreed for 2 reasons:
1. The equals method would have to be implemented in a production class (BigQueryDynamicTableSink) but is actually only used for testing. This becomes even worse when good coding practice dictate that if equals is overridden, then hashCode should also be implemented, resulting in even more unused production code that clutters the production class.
2. This class can evolve over time, with more attributes being added. We do not maintain the overhead of updating the equals and hashCode method every time this class's variables are modified.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
jayehwhyehentee: Good point. I'll raise a separate PR for that right away. Thank you

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
yerramalli: Yes, I did not call out overriding hashcode but the general practice is to override both and I don't see that as a bad thing. We never know how these objects are used in prod and in general it is a good practice to override both equals and hashcode.  But if you feel strongly about not implementing the equals and hashcode then you can convert it to a utility method with strongly typed arguments.

prashastia: ACK.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
prashastia: ACK, moved to the class.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
prashastia: ACK, moved to the class.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
prashastia: ACK, fixed.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkTest.java
prashastia: ACK, moved to the class.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSinkITCase.java
prashastia: Fixed.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
prashastia: Added, pls verify.

jayehwhyehentee (approved):

flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/fakes/StorageClientFaker.java
prashastia: I think the diff is causing confusion. 
Added a new method `createReadAndWriteOptions()` which initializes `FakeBigQueryServices` instance. `FakeBigQueryServices` has two params `FakeBigQueryStorageReadClient` and `FakeBigQueryStorageWriteClient`. `createTableReadOptions ()` sets `FakeBigQueryStorageWriteClient` to null as the read test does not require a `FakeBigQueryStorageWriteClient`. `createTableReadOptions()` were initially 4 methods with different signatures(depending upon the params provided, reduced them to 2 (the actually required method)) and added `createReadAndWriteOptions()` method for instantiating both `FakeBigQueryStorageReadClient` and `FakeBigQueryStorageWriteClient`

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactory.java
```diff
@@ -102,11 +113,32 @@ public DynamicTableSource createDynamicTableSource(Context context) {
             config = config.withTestingServices(testingServices);
         }
 
+        // Create an unbounded source.
```
clmccart, line 116: this comment is a little bit misleading. the boundedness of the source we create here is dependent on the config options (ie, won't always be unbounded)

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSink.java
```diff
@@ -0,0 +1,117 @@
...
+import com.google.cloud.flink.bigquery.sink.BigQuerySinkConfig;
+import com.google.cloud.flink.bigquery.sink.serializer.BigQuerySchemaProviderImpl;
+import com.google.cloud.flink.bigquery.sink.serializer.BigQueryTableSchemaProvider;
+import com.google.cloud.flink.bigquery.sink.serializer.RowDataToProtoSerializer;
+import org.apache.avro.Schema;
+
+import java.util.Objects;
+import java.util.ResourceBundle;
+
+/** A {@link org.apache.flink.table.connector.sink.DynamicTableSink} for Google BigQuery. Tho */
```
clmccart, line 36: extra "Tho"?

prashastia: Fixed!

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableFactory.java
prashastia: Fixed!

yerramalli (approved):

PR #146, opened 2024-08-08
fix: BQ partition utils intervals

amstee:
👋 been having an issue with the query restriction generated by the connector when running in unbounded mode.

The BETWEEN operator is inclusive: begin and end values are included.  this causes the source connector to read some of the data twice.

This change should fix that, let me know if there's any test I may have missed !


amstee:
@prashastia if you could review 🙏

jayehwhyehentee (approved):

amstee:
Hey, thanks for the review, any blocker @jayehwhyehentee ?

jayehwhyehentee:
@amstee merged

PR #163, opened 2024-09-26
imporve: tinyint should convert byte

caicancai:


caicancai:
@prashastia PATL, thanks

jayehwhyehentee:
@caicancai can you please add a unit test as well. Thanks!

caicancai:
> @caicancai can you please add a unit test as well. Thanks!

Sorry for the late reply. I was on vacation some time ago. This actually does not need to be tested, because it works fine with the original method. From the perspective of compiler principle, `element.getByte(fieldNumber)` is actually a subset of `(int) element.getShort(fieldNumber)`

caicancai:
@jayehwhyehentee By the way, I don't understand why it is forced to be converted to int

prashastia:
> > @caicancai can you please add a unit test as well. Thanks!
> 
> Sorry for the late reply. I was on vacation some time ago. This actually does not need to be tested, because it works fine with the original method. From the perspective of compiler principle, `element.getByte(fieldNumber)` is actually a subset of `(int) element.getShort(fieldNumber)`

 `getByte()` expects the value passed to be a byte. However, in the tests we explicitly pass `short` values in tests so that which needs to be modified.

prashastia:
> @jayehwhyehentee By the way, I don't understand why it is forced to be converted to int

The connector uses Storage Write API to write data to BigQuery. BQ only contains `INTEGER` type field which could store TINYINT, SMALLINT, INTEGER and BIGINT values. Storage Write API [datatype conversion](https://cloud.google.com/bigquery/docs/write-api#data_type_conversions) state that values to `INTEGER` fields should be atleast int32, which is why we cast even TINYINT and SMALLINT values to java `int`

caicancai:
thanks

PR #164, opened 2024-09-30
Improve debugability

prashastia:
1. For Datastream API, in case the source records' schema does not match the expected BQ Table(sink) schema, the deserializer (`BigQueryDeserializationSchema`) is unable to forward the mismatched record further to the pipeline. Hence, making this error more verbose to pinpoint the underlying cause instead of `ClassCastException` currently being seen due to the mismatch.

2. Fix error in writing time type to BigQuery. [LocalDateTime dropping 00 seconds value when parsing date string value with 00 seconds like "2018-07-06 00:00:00"](https://stackoverflow.com/questions/50786482/java-8-localdatetime-dropping-00-seconds-value-when-parsing-date-string-value-wi). Parsing now using ISO format to ensure a 00 at the end.

3. Fix unbounded read for time partitioning based on `DAY`, `MONTH` and `YEAR`
/gcbrun

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## Next
 
+* Support exactly-once consistency in BigQuery sink. This is integrated with Datastream and Table/SQL API.
+* Add Flink metrics for monitoring BigQuery sink.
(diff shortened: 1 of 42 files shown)


prashastia:
Combined with PR #176.

PR #178, opened 2024-10-28
Add integration tests for exactly-once

jayehwhyehentee:
/gcbrun


cloudbuild/nightly/cloudbuild.yaml
```diff
@@ -99,28 +99,10 @@ steps:
       - 'CLUSTER_TABLE_API_UNBOUNDED_TABLE_TEST_FILE=${_CLUSTER_TABLE_API_UNBOUNDED_TABLE_TEST_FILE}'
       - 'WORKER_MACHINE_TYPE_UNBOUNDED=${_WORKER_MACHINE_TYPE_UNBOUNDED}'
 
-# 4. Start the simple table e2e test.
```
prashastia, line 102: Any specific reason to remove the simple test. Shouldn't have atleast one simple test? In case nested check fails in future.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/BigQuerySource.java
```diff
@@ -351,14 +352,42 @@ public static BigQuerySource<GenericRecord> streamAvros(BigQueryReadOptions read
...
+                                                                connectOptions.getTable())),
                                                 tableSchema.getFields())
                                         .toString()))
                 .setReadOptions(readOptions)
                 .setSourceBoundedness(Boundedness.CONTINUOUS_UNBOUNDED)
                 .build();
     }
 
+    /** Replaces invalid characters with underscore. */
+    private static String sanitizeAvroSchemaName(String name) {
```
prashastia, line 365: Is there any specific case that fails or causes an issue?

cloudbuild/nightly/cloudbuild.yaml
```diff
@@ -281,8 +245,8 @@ steps:
       - 'CLUSTER_TABLE_API_UNBOUNDED_TABLE_TEST_FILE=${_CLUSTER_TABLE_API_UNBOUNDED_TABLE_TEST_FILE}'
       - 'SINK_PARALLELISM_UNBOUNDED_JOB=${_SINK_PARALLELISM_UNBOUNDED_JOB}'
 
-# Maximum tolerance 60 minutes.
-timeout: 3600s
+# Maximum tolerance 75 minutes.
+timeout: 4500s
```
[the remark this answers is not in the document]
jayehwhyehentee: 50 min

cloudbuild/nightly/cloudbuild.yaml
jayehwhyehentee: The simple table test was good while starting out, but now its redundant and doesnt add any value. If nested check fails in the future then we have to debug anyway.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/BigQuerySource.java
jayehwhyehentee: If source table name has hyphen, then avro schema name validation will fail. It only accepts alphabets, digits and underscore. Found this while fixing the BQ table names in our e2e tests.

jayehwhyehentee: Since the avro schema name is completely internal and hidden from the user, it doesnt matter how we sanitize. We just need to make sure that a valid BQ table name is not rejected by the connector because of some internal avro standard.

jayehwhyehentee: Note that this is on the source side only.

prashastia (approved):

PR #182, opened 2024-11-12
Fix snapshot metrics

prashastia:
Shift since snapshot metrics to `snapshotState()` rather than first write to prevent leaking the metric values to scope of the next checkpoint.

/gcbrun


jayehwhyehentee (changes requested):
In the tests, instead of checking if "since checkpoint" counts are zero after "snapshotState", move those checks to a point before "snapshotState" where they have meaningful values.

Presently, with a trivial zero check, tests for those metrics are useless.

jayehwhyehentee (approved):

Issue #185, opened 2024-11-19
Dropwizard Dependency Can Cause Job Failures


byronellis:
The DropWizard metrics plugin is not shipped by default with Flink so will cause failed jobs when you try to use it with a "stock" Flink installation. Worse yet this error will not be surfaced unless you explicitly set the job type of BATCH.

PR #186, opened 2024-11-20
remove dropwizard dependency from bq source

clmccart:
For context on why we want to remove this dep:  #185 

This means that we will no longer record the `readSplitTime` histogram metric. We believe this is a reasonable tradeoff since read split times are still logged and read times can be monitored on the service side.


clmccart:
@jayehwhyehentee PTAL whenever you get a chance!

jayehwhyehentee (approved):

PR #192, opened 2025-01-06
Support table creation in sink

jayehwhyehentee:
- Use avro schema in BigQuerySchemaProvider to derive BigQuery table schema
- Use Flink's in-built logical type to avro schema converter for Table API
- Change throttling logic to accommodate table creation
- Rely on default BigQuery client retries when creating BigQuery table

Also, remove duplicate connector exception class from sink

/gcbrun


flink-1.17-connector-bigquery/flink-connector-bigquery/src/test/java/com/google/cloud/flink/bigquery/source/split/reader/deserializer/AvroToRowDataConvertersTest.java
```diff
@@ -268,6 +269,7 @@ record =
         Assertions.assertThat(exception).hasMessageContaining("Avro to RowData Conversion Error");
     }
 
+    @Ignore
```
jayehwhyehentee, line 272: @prashastia need to fix this test before release. I have disabled it for now.

prashastia (approved):

PR #207, opened 2025-02-05
Apply constant traceID in default writer

jayehwhyehentee:
- The connector uses connection pool in default writer as it ensures more efficient usage of the default stream.
- Reason for this commit is a limitation in BQ storage write API's connection pool feature.
- If a user has multiple writers writing to BQ tables in the same BQ region, then all connections must have same traceID.

/gcbrun


prashastia (approved):

agrawal-siddharth (approved):
Note that we do plan to eventually remove this BQ API limitation, at which time you can revert back to your original solution.

Issue #213, opened 2025-02-19
Parametrise BigQuerySink


mina-asham:
Currently the generic type is not propagated to `com.google.cloud.flink.bigquery.sink.BigQuerySink#get` which makes the utility generate a non-generic Sink giving warnings in projects using that, but also making it hard to discover some bugs due to typing missing for tools and IDEs, we should support that

jayehwhyehentee:
Thanks for raising this issue. Proposed solution at https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/218. Unfortunately this repository has not been opened for public contributions yet, so we cannot merge your commit.

PR #214, opened 2025-02-19
fix: Parametrize BigQuerySink with generics

mina-asham:
- Generics where already used in the underlying level but not at the final level, so just propagate that along the way
- Useful for early error detection and for general tooling and detection
- Fixes: https://github.com/GoogleCloudDataproc/flink-bigquery-connector/issues/213


jayehwhyehentee:
As mentioned in associated issue, this repo has not been opened for public contributions yet. Kindly check https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/218 for this fix.

PR #224, opened 2025-04-10
fix: Fix shaded jar relocations

nickcaballero:
This PR excludes the Flink modules from the shaded artifact and ensures that the Flink SPIs are not relocated.


yalimu-g:
Close it as it is addressed in PR #260

PR #295, opened 2026-03-19
Bigquery upsert support for flink 2.1

mhynes-shopify:
## Summary
- Add upsert (changelog) support to bigquery connector for flink 2.1
- Based on https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/262

### DIFF
--- a/flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/BigQueryProtoSerializer.java
+++ b/flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/BigQueryProtoSerializer.java
@@ -74,8 +74,9 @@ public abstract class BigQueryProtoSerializer<IN> implements Serializable {
     /**
      * Convert Flink record to proto ByteString with CDC metadata included.
      *
-     * <p>The default implementation calls {@link #serialize(Object)}. Subclasses that support CDC
-     * should override this method to include the CDC pseudocolumns in the serialized output.
+     * <p>The default implementation throws {@link UnsupportedOperationException}. Subclasses that
+     * support CDC must override this method to include the CDC pseudocolumns in the serialized
+     * output.
      *
      * @param record Record to serialize.
      * @param changeType The CDC change type ("UPSERT" or "DELETE").
@@ -85,9 +86,11 @@ public abstract class BigQueryProtoSerializer<IN> implements Serializable {
      */
     public ByteString serializeWithCdc(IN record, String changeType, String changeSequenceNumber)
             throws BigQuerySerializationException {
-        // Default implementation ignores CDC fields for backward compatibility.
-        // Subclasses should override to include CDC pseudocolumns.
-        return serialize(record);
+        throw new UnsupportedOperationException(
+                String.format(
+                        "Serializer %s does not support CDC serialization. "
+                                + "Override serializeWithCdc() to handle change type '%s'.",
+                        getClass().getName(), changeType));
     }
 
     /**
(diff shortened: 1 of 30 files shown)


flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/BigQueryProtoSerializer.java
```diff
@@ -56,4 +56,54 @@ public void init(BigQuerySchemaProvider schemaProvider) {}
...
+     *
+     * <p>The default implementation returns null. Subclasses should override this method to extract
+     * a meaningful sequence number from the record.
+     *
+     * @param record Record to extract sequence number from.
+     * @param sequenceField The name of the field containing the sequence value, or null.
+     * @return Hexadecimal string representation of the sequence number, or null to omit.
+     */
+    public String extractSequenceNumber(IN record, String sequenceField) {
+        return null;
```
yalimu-g: Is this intended to return null?

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
```diff
@@ -197,8 +212,17 @@ void ensureSchemaAwareness(IN record) {
...
             }
         }
-        protoSchema = getProtoSchema(schemaProvider);
-        serializer.init(schemaProvider);
+
+        if (cdcEnabled) {
+            BigQuerySchemaProvider cdcSchemaProvider =
+                    new BigQueryCdcSchemaProvider(schemaProvider);
+            protoSchema = getProtoSchema(cdcSchemaProvider);
+            serializer.init(schemaProvider);
```
yalimu-g: this seems to be the same for both cdc enabled or not?

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BigQueryBufferedWriter.java
```diff
@@ -153,7 +153,10 @@ public BigQueryBufferedWriter(
                 createTableOptions,
                 fatalizeSerializer,
                 maxParallelism,
-                traceId);
+                traceId,
+                false, // CDC not supported with exactly-once (uses buffered streams)
```
yalimu-g: nit: add comments of the parameters for the default value

yalimu-g (approved):

yalimu-g (approved):

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/serializer/BigQueryProtoSerializer.java
mhynes-shopify: That's a good Q, thanks. This is the base class and the child classes should provide the concrete implementation; I updated it to throw an exception instead

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/config/BigQueryConnectorOptions.java
```diff
@@ -233,4 +233,52 @@ private BigQueryConnectorOptions() {}
...
+                    .noDefaultValue()
+                    .withDescription(
+                            "Comma-separated list of primary key columns to apply when the connector "
+                                    + "auto-creates a CDC destination table.");
+
+    /**
+     * [OPTIONAL, Sink Configuration] BigQuery INTERVAL literal used as max_staleness when
+     * auto-creating a CDC destination table.
+     */
+    public static final ConfigOption<String> CDC_MAX_STALENESS =
```
mhynes-shopify, line 285: @yalimu-g  I did find there was something strange about this:
- when I tested this the table creation succeeded, but the table didn't actually have the max staleness set
- the default value was not configured, so we essentially have the staleness unset

I don't know why the staleness wasn't accepted, but how the create still succeeded. After creation I was still able to run `alter table` to set the max staleness.

An unset max staleness isn't a blocker for us, but it would be strange to let this silently pass through. I think we should at least document that it doesn't work presently

yalimu-g: It seems this is an known issue for the Bigquery rest api. I would recommend either remove it from here or add a follow up todo for alter the table after table creation is this is set.

mhynes-shopify: SG thanks for taking a look; will document it for now

mhynes-shopify: https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/295/changes/118bc1369790a3d6d1bbd99dd848c62d9c1adb68

PR #305, opened 2026-04-27
Fix critical data loss bug: signal no-more-splits per reader and remove readers from queue 

fqshopify:
TLDR; Two (race condition) bugs in BigQuerySourceEnumerator.assignSplits compound to produce silent row loss:

- Bug 1 fails already-finishing reader tasks and returns their splits to the pool.
- Bug 2 then reassigns those returned splits to reader tasks that have since terminated. The reassigned splits sit booked to dead tasks and are never re-read.

Actual code changes is only ~10 lines. Rest is just docs. 

Note: the docs in the PR are slightly wrong and were removed in a later PR. See [this](https://github.com/user-attachments/files/27139875/pr_description.html) for an indepth and accurate explanation of how the bug manifests.

### DIFF
--- a/docs/pr_description.md
+++ b/docs/pr_description.md
@@ -1,0 +1,428 @@
+# Background
+
+Terms used throughout:
+
+- **Split** - a unit of input work wrapping one BigQuery Storage API *stream*. The stream yields a
+  subset of the source table's rows; reading the split streams those rows into Flink. Row counts per
+  stream vary. The enumerator hands splits to readers; each reader reads its splits sequentially. In
+  these runs the table is partitioned into 3,000 streams → 3,000 splits.
+- **Reader / registered reader** - a source subtask that has registered with the JobManager and will
+  request splits.
+- **Enumerator** - lives on the JobManager and hands out splits to readers on demand.
+- **`NoMoreSplitsEvent`** - event the enumerator signals when the assigner is drained; tells a
+  reader it can finish once its current split is done.
+
+# What?
+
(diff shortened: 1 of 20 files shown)


docs/pr_description.md
fqshopify, line 1: Will remove the files under `docs/` and move them into the PR description before merging. 

Putting it here for now to make conversations easier.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/enumerator/BigQuerySourceEnumerator.java
```diff
@@ -136,8 +140,9 @@ private void assignSplits() {
                 LOG.info("Assign split {} to subtask {}", bqSplit, nextAwaiting);
                 break;
             } else if (splitAssigner.noMoreSplits() && boundedness == Boundedness.BOUNDED) {
-                LOG.info("All splits have been assigned");
-                context.registeredReaders().keySet().forEach(context::signalNoMoreSplits);
+                LOG.info("Signal NoMoreSplits to subtask {}", nextAwaiting);
+                context.signalNoMoreSplits(nextAwaiting);
+                awaitingReader.remove();
```
fqshopify, line 145: This is the actual fix; only 3 lines changes.

yalimu-g (approved):

docs/pr_description.md
fqshopify: Following up on this here: https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/306

PR #307, opened 2026-04-28
Bq view flink

bipinupd:
Changes:

- BigQueryServices & BigQueryServicesImpl: Added isView and materializeView methods to handle interaction with the BigQuery backend.
- SplitDiscoverer: Updated to check if the target resource is a view. If it is, and the feature is enabled, it triggers the materialization process and routes the subsequent Storage Read Session to the newly created temporary table.

Verification

- All unit tests in flink-connector-bigquery-common passed successfully.
- Verified that the integration test package builds successfully with the new configurations.


flink-connector-bigquery-common/src/main/java/com/google/cloud/flink/bigquery/common/config/BigQueryConnectOptions.java
```diff
@@ -88,15 +92,19 @@ public final boolean equals(Object obj) {
...
+                .setMaterializedTableExpirationHours(24);
     }
 
     public static Builder builderForQuerySource() {
         return new AutoValue_BigQueryConnectOptions.Builder()
                 .setCredentialsOptions(CredentialsOptions.builder().build())
                 .setProjectId("")
                 .setDataset("")
-                .setTable("");
+                .setTable("")
```
d-cole, line 117: Similar to the spark connector, it would be very useful to configure both of these independent of the project containing the view definition,
 - Project where the temp table is persisted (materializationProject & materializationDataset in spark connector)
 - Project where compute to materialize the table is run/billed (billingProject in spark connector)

bipinupd: Thanks @d-cole. I have added option for temp dataset persistance and billing isolation.

bipinupd:
@yalimu-g Can please review this PR?

yalimu-g (approved):

PR #310, opened 2026-05-04
Indirect writes (2/8): BigQueryLoadJobOperator

fqshopify:
Adds three methods to `BigQueryServices.QueryDataClient` (`submitJob`, `getJob`, `waitForJob`) and a new `BigQueryLoadJobOperator` that will run in the post-commit topology of the indirect-write sink. The operator aggregates committed GCS file paths per checkpoint, submits BigQuery load jobs with deterministic job IDs, and cleans up the staged GCS files after the load completes.

`CheckpointState` is the per-checkpoint tracking helper used by the operator.

Includes `StorageClientFaker` stubs for both 1.17 and 2.1 trees so the extended `QueryDataClient` interface still compiles in tests.

--- 

This is (2/8) in the indirect-writes series:

- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/309
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/310
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/311
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/312
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/313
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/317
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/318
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/319

PR #309, opened 2026-05-04
Indirect writes (1/8): Design doc

fqshopify:
Adds the design doc for the upcoming indirect-write sink (BigQuery batch load-job pipeline that writes Parquet to GCS, then issues a load job into BigQuery) under `docs/BigQuery Indirect Writes Sink.md`.

This is mostly provided as a reference and is independent of the rest of the PRs. Can merge it (or not) any time, none of the other PRs depend on it. 

- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/309
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/310
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/311
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/312
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/313
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/317
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/318
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/319

PR #311, opened 2026-05-04
Indirect writes (3/8): BigQueryIndirectSink

fqshopify:
Adds `BigQueryIndirectSink`, a `Sink<IN>` that writes data as Parquet to GCS via Flink's `FileSink` and then dispatches `BigQueryLoadJobOperator` in the post-commit topology to load the staged files into BigQuery. Provides `EXACTLY_ONCE` delivery.

New supporting classes:
- `WriteMode` enum `{STORAGE_WRITE_API, INDIRECT}`
- `RowDataParquetWriterFactory`: `ParquetWriterFactory<RowData>` with INT64 timestamp micros encoding so BigQuery `TIMESTAMP`s round-trip through Parquet with microsecond precision
- `SizeBasedCheckpointRollingPolicy`: rolls files at 1.5 GB

`BigQuerySink` dispatches to `BigQueryIndirectSink` when `config.getWriteMode() == INDIRECT`. `BigQuerySinkConfig` adds `writeMode`, `gcsTempPath`, and `bulkWriterFactory` fields plus `validateIndirect()`.

Pulls in `flink-parquet`, `hadoop-common 2.10.2`, and `hadoop-mapreduce-client-core 2.10.2`. The `commons-io 2.15.1` test override is forced because Hadoop 2.10 transitively brings `commons-io 2.4`, which lacks `NullPrintStream` that Flink's `MiniCluster` needs at test time.

---

This is (3/8) in the indirect-writes series:

- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/309
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/310
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/311
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/312
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/313
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/317
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/318
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/319

### DIFF
--- a/flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
+++ b/flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (C) 2024 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.flink.bigquery.sink;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.RuntimeExecutionMode;
+import org.apache.flink.api.connector.sink2.Committer;
+import org.apache.flink.api.connector.sink2.CommitterInitContext;
+import org.apache.flink.api.connector.sink2.Sink;
+import org.apache.flink.api.connector.sink2.SinkWriter;
+import org.apache.flink.api.connector.sink2.StatefulSinkWriter;
+import org.apache.flink.api.connector.sink2.SupportsCommitter;
+import org.apache.flink.api.connector.sink2.SupportsWriterState;
+import org.apache.flink.api.connector.sink2.WriterInitContext;
+import org.apache.flink.configuration.ExecutionOptions;
+import org.apache.flink.connector.file.sink.FileSink;
+import org.apache.flink.connector.file.sink.FileSinkCommittable;
+import org.apache.flink.connector.file.sink.writer.FileWriterBucketState;
+import org.apache.flink.core.fs.Path;
+import org.apache.flink.core.io.SimpleVersionedSerializer;
+import org.apache.flink.streaming.api.connector.sink2.CommittableMessage;
+import org.apache.flink.streaming.api.connector.sink2.SupportsPostCommitTopology;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.streaming.api.functions.sink.filesystem.bucketassigners.BasePathBucketAssigner;
+
+import com.google.cloud.bigquery.FormatOptions;
+import com.google.cloud.flink.bigquery.common.config.BigQueryConnectOptions;
+import com.google.cloud.flink.bigquery.sink.indirect.BigQueryLoadJobOperator;
+import com.google.cloud.flink.bigquery.sink.indirect.SizeBasedCheckpointRollingPolicy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.UUID;
+
+/**
+ * BigQuery sink using indirect writes (GCS files + load jobs).
+ *
+ * <p>This sink writes data as Parquet files to GCS using Flink's {@link FileSink} and then submits
+ * BigQuery load jobs in the post-commit topology to load the data into the destination table.
+ *
+ * <p>The pipeline is:
+ *
+ * <pre>
+ * RowData
+ *   -&gt; FileSink's FileWriter (BulkWriter.Factory uses RowDataParquetWriterFactory)
+ *   -&gt; [Checkpoint]
+ *   -&gt; FileSink's FileCommitter (finalizes .inprogress -&gt; final GCS files)
+ *   -&gt; PostCommitTopology: BigQueryLoadJobOperator
+ *        -&gt; Aggregates committed files per (subtask, checkpoint)
+ *        -&gt; Submits BigQuery load job(s) with deterministic job IDs
+ *        -&gt; Cleans up GCS files
+ * </pre>
+ *
+ * <p>The sink provides EXACTLY_ONCE delivery guarantee.
+ *
+ * <p>Requires the GCS Hadoop connector plugin ({@code flink-gs-fs-hadoop}) to be installed in the
+ * Flink cluster's {@code plugins/} directory.
+ */
+@Internal
+final class BigQueryIndirectSink<IN>
+        implements Sink<IN>,
+                SupportsWriterState<IN, FileWriterBucketState>,
+                SupportsCommitter<FileSinkCommittable>,
+                SupportsPostCommitTopology<FileSinkCommittable> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BigQueryIndirectSink.class);
+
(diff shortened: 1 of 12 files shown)


flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Override
+    public void addPostCommitTopology(
+            final DataStream<CommittableMessage<FileSinkCommittable>> committables) {
+        if (committables
+                        .getExecutionEnvironment()
+                        .getConfiguration()
+                        .get(ExecutionOptions.RUNTIME_MODE)
+                != RuntimeExecutionMode.BATCH) {
+            throw new IllegalStateException(
```
drossos, line 153: I thought from the original spec streaming mode was supported as well

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
```diff
@@ -0,0 +1,160 @@
...
+                    "INDIRECT write mode is only supported in BATCH execution mode");
+        }
+
+        committables
```
drossos, line 157: Lol I was wondering how in part 2 of PR how loadjoboperator was going to handle pllsm. Totally makes sense now

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
drossos: ref: https://github.com/Shopify/flink-bigquery-connector/blob/11133c07736f5ca51ebf8c82e643235412126a7a/docs/BigQuery%20Indirect%20Writes%20Sink.md#streaming-mode-flow

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
```diff
@@ -0,0 +1,160 @@
...
+                .global()
```
drossos, line 158: I understand that the BigQueryLoadJobOperator is set to pllsm of 1 and all on the subtask ensuring that every committer message comes through that singular flat-map instance. Inside the flatmap we then block on checkpoint completion + load-job in BQ.

I am wondering if you have explored / tested alternatively keying the stream out after the "FileSink's FileCommitter" (or maybe even at the FileWriter) to allow for a deterministic breakup of the CommitableMessages that would still support the same on-checkpoint completion mechanism for writing out to BQ. 

I am pretty shakey on if this would be a valid alternative / would even reap significant performance gains so might be better to discuss over a call.

drossos, line 158: Originally was going to point / ask if unaligned checkpoints would ruin this design, but upon investigation realized that Flink disables UC for sink post-commit edges as a policy meaning we don't have to worry about weird behaviours with checkpoint barriers.

This might be worth noting down in design or comment somewhere.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/RowDataParquetWriterFactory.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Objects;
+
+/**
+ * Factory for Parquet bulk writers used by the indirect-write path.
+ *
+ * <p>Configures Parquet to encode timestamps as INT64 with microsecond precision so BigQuery's
+ * Parquet load reads them back as {@code TIMESTAMP} with full TIMESTAMP(6) fidelity. The default
+ * INT96 encoding would lose precision and is treated as a legacy Hive type by BigQuery.
+ *
+ * <p><b>Why this extends {@code ParquetWriterFactory} instead of returning one directly:</b> {@code
```
drossos: I am still a bit unclear on on why we have to extend the existing ParquetWriterFactory.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
fqshopify: Correct; the design/spec does show how we can support streaming mode. 

For the initial implementation/PRs though, I'm focusing only on batch mode (that's our primary use case for this feature at the moment). 

Streaming needs pretty rigorous testing (and is harder to change once it's released) so it's better to deal with that separately.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
fqshopify: Yes I considered this but here's the main problem; BigQuery enforces a 1,500 load-job-limit-per-table-per-day. 
So you really want to minimize the number of load jobs you do a day per table.
If you start keying by something, you're increasing the number of load jobs. 

For example, a reasonable idea might be to key by subtaskId so you have one load job per writer essentially. 
Now imagine you have a large batch job (think >1000 parallelism). 
You would be submitting 1000 load jobs her Flink batch job run. 
Which means you could run that batch job at most once per day. 

In conclusion, I saw no benefits to keying this operation and only downsides.

fqshopify: Good point, I'll add that to the design doc. 

For batch though, this is irrelevant (no real checkpointing in batch mode).

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/RowDataParquetWriterFactory.java
fqshopify: Existing `ParquetWriterFactory` doesn't implement equals/hashCode. 
My `RowDataParquetWriterFactory` does implement equals/hashCode.

This type is carried around by BigQuerySinkConfig which i do want to ideally have proper value-equality. 

I'm going to remove this confusing javadoc comment and just add a proper unit test for this.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
fqshopify: Done

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/RowDataParquetWriterFactory.java
fqshopify: Done

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/BigQueryIndirectSink.java
drossos: Totally makes sense and in agreement 👍

drossos (approved):
https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/309#pullrequestreview-4253932845

yalimu-g (approved):
Thanks for contributing!

fqshopify:
rebased on top of main!

yalimu-g (approved):

PR #316, opened 2026-05-12
Nested projection pushdown support

drossos:
Enables nested projection pushdown for the BigQuery connector by flipping `supportsNestedProjection()` to `true`. This will then cause our generated plans from the Flink table planner to include projected fields in its projection planning.

** Source Side**
This table plan is represented as `int[][] projectedFields` where `projectedFields[i]` represents the column, and `projectedFields[i][j]` represents the field index at level `j` in the row tree. We then build `String[][] fullyQualifiedPaths` which represents the same thing as projectedFields, but shows the full names of the path to the projected value. (This will then be saved and used later in the deserialization phase). Example mapping being `[0,1,0]` → `["payload","buyer","id"]`. Using this fullyQualifiedPaths, we then 'dot' separate and pass to BigQuery Storage API as `selected_fields` when creating a read session.            

** Deserializer Side**
We then get Avro records back with all fields from the projection in a nested structure. On the first record, as before we call `computeIndexMapping` to resolve where in each Avro message our field names exist. The only main difference is now on `createRowConverter` we can pass an optional arg `nestedPaths`. When present, `computeIndexMapping` walks each path through the Avro schema's nested records (transparently unwrapping ["null", T] nullable unions) and resolves it to an int[] of field positions per output column. Each path can either reach a leaf (returning a scalar via the appropriate converter) or stop at a parent struct, in which case BQ returns the whole sub-record and the converter walks it via its existing recursive ROW case.

*Note*: Chose to use the fully qualified names instead of indices on our deserialization because there isn't any explicit guarantee on our Avro field ordering we get back, so by using the qualified names (similar to how it was done before in flat case), we ensure we are indexing only after obtaining via field name.



*Top-hatting* 

``` java
/**
 * Schema:
 * id        INTEGER  NULLABLE
 * name      STRING   NULLABLE
 * address   RECORD   NULLABLE
 *   city      STRING   NULLABLE
 *   country   RECORD   NULLABLE
 *     code      STRING   NULLABLE
 *     name      STRING   NULLABLE
 *   current   BOOLEAN  NULLABLE
 */
  public static void main(String[] args) {                                                                                                                                                                                         
      var env = StreamExecutionEnvironment.createLocalEnvironment(new Configuration());                                                                                                                                            
      var t = StreamTableEnvironment.create(env);                                                                                                                                                                                  
                                                                                                                                                                                                                                   
      //   id BIGINT, name STRING,                                                                                                                                                                                                 
      //   address ROW<city STRING, country ROW<code STRING, name STRING>, `current` BOOLEAN>                                                                                                                                      
      t.executeSql(                                                                                                                                                                                                                
          "CREATE TABLE nested_tbl ("
              + "  id BIGINT, name STRING,"                                                                                                                                                                                        
              + "  address ROW<city STRING, country ROW<code STRING, name STRING>, `current` BOOLEAN>"                                                                                                                             
              + ") WITH ('connector'='bigquery',"                                                                                                                                                                                  
              + "        'project'='xxx',"                                                                                                                                                                         
              + "        'dataset'='xxx',"                                                                                                                                                                                      
              + "        'table'='xxx')");          

      // Mix: top-level scalar + 1-level nested leaf + 2-level nested leaf                                                                                                                                                         
      t.executeSql("EXPLAIN SELECT id, name, address.city, address.country.code FROM nested_tbl LIMIT 5").print();
      t.executeSql("SELECT id, name, address.city, address.country.code FROM nested_tbl LIMIT 5").print();                                                                                                                         
  } 
```

### DIFF
--- a/flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSource.java
+++ b/flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSource.java
@@ -33,6 +33,7 @@ import org.apache.flink.table.connector.source.abilities.SupportsProjectionPushD
 import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.expressions.ResolvedExpression;
 import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.logical.LogicalType;
 import org.apache.flink.table.types.logical.RowType;
 
 import com.google.cloud.flink.bigquery.common.config.BigQueryConnectOptions;
@@ -48,6 +49,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -69,16 +71,41 @@ public class BigQueryDynamicTableSource
     private BigQueryReadOptions readOptions;
     private DataType producedDataType;
 
-    public BigQueryDynamicTableSource(BigQueryReadOptions readOptions, DataType producedDataType) {
-        if (readOptions.getColumnNames().isEmpty()) {
-            readOptions =
-                    readOptions
-                            .toBuilder()
-                            .setColumnNames(DataType.getFieldNames(producedDataType))
-                            .build();
-        }
+    private final RowType physicalDataType;
+    private String[][] fullyQualifiedPaths;
+
+    public BigQueryDynamicTableSource(BigQueryReadOptions readOptions, DataType physicalDataType) {
+        this(
+                ensureDefaultColumnNames(readOptions, physicalDataType),
+                (RowType) physicalDataType.getLogicalType(),
+                physicalDataType,
+                defaultFlatPaths((RowType) physicalDataType.getLogicalType()));
+    }
+
+    private BigQueryDynamicTableSource(
+            BigQueryReadOptions readOptions,
+            RowType physicalDataType,
+            DataType producedDataType,
+            String[][] fullyQualifiedPaths) {
         this.readOptions = readOptions;
+        this.physicalDataType = physicalDataType;
         this.producedDataType = producedDataType;
+        this.fullyQualifiedPaths = fullyQualifiedPaths;
+    }
+
+    private static BigQueryReadOptions ensureDefaultColumnNames(
+            BigQueryReadOptions readOptions, DataType physicalDataType) {
+        if (readOptions.getColumnNames().isEmpty()) {
+            return readOptions
+                    .toBuilder()
+                    .setColumnNames(DataType.getFieldNames(physicalDataType))
+                    .build();
+        }
+        return readOptions;
+    }
+
+    private static String[][] defaultFlatPaths(RowType rowType) {
+        return rowType.getFieldNames().stream().map(f -> new String[] {f}).toArray(String[][]::new);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSource.java
```diff
@@ -116,19 +146,48 @@ public String asSummaryString() {
...
+        return Arrays.stream(paths)
+                .map(indexPath -> toNamePath(indexPath, this.physicalDataType))
+                .toArray(String[][]::new);
+    }
+
+    private static String[] toNamePath(final int[] indexPath, final RowType rowType) {
+        final String[] namePath = new String[indexPath.length];
+        LogicalType currentType = rowType;
+        for (int i = 0; i < indexPath.length; i++) {
+            final RowType row = (RowType) currentType;
```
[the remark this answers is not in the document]
drossos: SupportsProjectionPushDown.supportsNestedProjection() is by contract only invoked to descend through ROW types. Flink planner would have to introduce breaking change for this behaviour to change so should operate based on the contract we have now

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/source/reader/deserializer/AvroToRowDataConverters.java
```diff
@@ -88,53 +88,85 @@ public interface AvroToRowDataConverter extends Serializable {
...
 
+    // Check if nullable and take non-null path for schema unwrap
+    private static org.apache.avro.Schema unwrapNullableUnion(org.apache.avro.Schema schema) {
+        if (!schema.isUnion()) {
+            return schema;
+        }
+        return schema.getTypes().stream()
+                .filter(s -> s.getType() != org.apache.avro.Schema.Type.NULL)
+                .findFirst()
+                .orElseThrow();
```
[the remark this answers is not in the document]
drossos: Made them consistent with suggestion

drossos:
Friendly CC to @yalimu-g for review

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSource.java
```diff
@@ -116,19 +146,48 @@ public String asSummaryString() {
 
     @Override
     public boolean supportsNestedProjection() {
-        return false;
+        return true;
```
BoqianShi, line 149: Could you add a Table/SQL test that selects nested fields end-to-end and verifies both the BigQuery dotted `selected_fields` and the final flattened `RowData`, since this PR changes the planner/source/deserializer contract rather than only the local converter logic? Same concern applies to the 2.1 connector copy.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSource.java
BoqianShi, line 176: Since `supportsNestedProjection()` now advertises nested projection support broadly, could you either support or explicitly reject paths through repeated RECORD fields / `ARRAY<ROW>` with tests, because this implementation assumes every intermediate path component is a `RowType` and will otherwise fail with an opaque cast error? Same concern applies to the 2.1 connector copy.

drossos: Hey Boqian, we also have a conversation about this internally and would love to hear your two-cents. 

`supportsNestedProjection()` contract comes from Flink and guarentees that they will never give you a projection where there are non-row types to descend through. If they do that is a contract break from the Flink side and should be a breaking change to our connector.

BoqianShi: Thanks, that makes sense. I checked the Flink projection utility as well. The nested paths passed to `applyProjection` are produced from `RexFieldAccess` chains, while array/item access does not get pushed as a deeper path through the array element. So this should not receive a path that requires descending through a non-ROW intermediate type under the current Flink planner behavior. Resolving this concern.

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/BigQueryDynamicTableSource.java
drossos: Added new integration test + modified faker client to allow for checking BigQuery read session is configured with the dotted selected_fields

fqshopify:
CI passed

BoqianShi (approved):
Thanks, this addresses my requested coverage.

Looks good to me.

PR #317, opened 2026-05-12
Indirect writes (6/8): large-workload (>15 TB / >10K files) support

fqshopify:
Partition each checkpoint's files when it exceeds BigQuery's per-load-job caps (10K source URIs or 15 TB), load each partition into a temp table, then atomically COPY all temp tables into the destination. Files are sorted by GCS path before partitioning so retries produce the same partition shape — without that, a retry could place a file into a different temp-table slot than the original attempt did, and since temp-load jobs are idempotent on jobId (not content) the COPY would double-count it.

Temp-load jobs for the partitions are submitted in parallel via the executor service; the COPY job still waits for all of them. Failures on any partition surface immediately while siblings keep running — those keep-running jobs aren't wasted because temp loads are idempotent on jobId, so a Flink retry re-attaches to any in-flight or completed BQ jobs without re-uploading.

Adds a required `write.indirect.temp-bigquery-dataset` option pointing at a separate dataset for temp tables. The recommendation is to set a default `tableExpirationMs` on that dataset so BigQuery itself cleans up any temp tables left behind by failed jobs.

---

This is (6/8) in the indirect-writes series:

- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/309
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/310
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/311
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/312
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/313
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/317
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/318
- https://github.com/GoogleCloudDataproc/flink-bigquery-connector/pull/319

### DIFF
--- a/flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
+++ b/flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
@@ -24,10 +24,11 @@ import org.apache.flink.connector.file.sink.FileSinkCommittable;
 import org.apache.flink.core.fs.FileSystem;
 import org.apache.flink.core.fs.Path;
 import org.apache.flink.streaming.api.connector.sink2.CommittableMessage;
-import org.apache.flink.streaming.api.functions.sink.filesystem.InProgressFileWriter;
+import org.apache.flink.streaming.api.functions.sink.filesystem.InProgressFileWriter.PendingFileRecoverable;
 import org.apache.flink.util.Collector;
 import org.apache.flink.util.function.SerializableSupplier;
 
+import com.google.cloud.bigquery.CopyJobConfiguration;
 import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobConfiguration;
@@ -42,6 +43,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -67,11 +70,40 @@ public class BigQueryLoadJobOperator
     private static final long EXECUTOR_SHUTDOWN_TIMEOUT_SECONDS = 10L;
     private static final ThreadFactory EXECUTOR_THREAD_FACTORY = r -> new Thread(r, "bq-load");
 
+    /**
+     * Default maximum source URIs per BigQuery load job. See <a
+     * href="https://cloud.google.com/bigquery/quotas#load_jobs">BigQuery load job quotas</a>.
+     */
+    static final int DEFAULT_MAX_FILES_PER_LOAD_JOB = 10_000;
+
+    /**
+     * Default maximum data per BigQuery load job (15 TB). See <a
+     * href="https://cloud.google.com/bigquery/quotas#load_jobs">BigQuery load job quotas</a>.
+     */
+    static final long DEFAULT_MAX_BYTES_PER_LOAD_JOB = 15L * 1024 * 1024 * 1024 * 1024;
+
+    /**
+     * Format for per-checkpoint job/table IDs: {@code <prefix>_<jobIdHex>_<uuid>_c<checkpointId>}.
+     */
+    private static final String CHECKPOINT_ID_FORMAT = "%s_%s_%s_c%d";
+
+    /** Per-partition variant of {@link #CHECKPOINT_ID_FORMAT}: appends {@code _p<idx>}. */
+    private static final String PARTITION_ID_FORMAT = CHECKPOINT_ID_FORMAT + "_p%d";
+
+    private static final String LOAD_JOB_ID_PREFIX = "flink-bq-load";
+    private static final String TEMP_TABLE_NAME_PREFIX = "flink-bq-tmp";
+    private static final String TEMP_LOAD_JOB_ID_PREFIX = "flink-bq-tmp-load";
+    private static final String COPY_JOB_ID_PREFIX = "flink-bq-copy";
+
     private final BigQueryConnectOptions connectOptions;
     private final UUID uuid;
+    private final String tempProject;
+    private final String tempDataset;
     private final SerializableSupplier<ExecutorService> executorServiceSupplier;
     private final SerializableSupplier<FileSystem> fileSystemSupplier;
     private final FormatOptions formatOptions;
+    private final int maxFilesPerLoadJob;
+    private final long maxBytesPerLoadJob;
 
     private transient BigQueryServices.QueryDataClient queryClient;
     private transient String jobIdHex;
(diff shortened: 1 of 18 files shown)


flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
```diff
@@ -66,10 +70,38 @@ public class BigQueryLoadJobOperator
...
+     */
+    static final long DEFAULT_MAX_BYTES_PER_LOAD_JOB = 15L * 1024 * 1024 * 1024 * 1024;
+
+    /**
+     * Format for per-checkpoint job/table IDs: {@code <prefix>_<jobIdHex>_<uuid>_c<checkpointId>}.
+     */
+    private static final String CHECKPOINT_ID_FORMAT = "%s_%s_%s_c%d";
+
+    /** Per-partition variant of {@link #CHECKPOINT_ID_FORMAT}: appends {@code _p<idx>}. */
+    private static final String PARTITION_ID_FORMAT = "%s_%s_%s_c%d_p%d";
```
drossos: nit: If we want to enforce partition id is dependent on checkpoint id, should this then be the checkpoint id appended on? e.g `PARTITION_ID_FORMAT = CHECKPOINT_ID_FORMAT+"_p%d";`

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
```diff
@@ -147,71 +191,226 @@ public void flatMap(
...
+     */
+    private List<List<PendingFileRecoverable>> partitionFiles(
+            final List<PendingFileRecoverable> files) {
+
+        files.stream()
+                .filter(file -> file.getSize() > maxBytesPerLoadJob)
+                .findFirst()
+                .ifPresent(
+                        file -> {
+                            throw new BigQueryConnectorException(
```
drossos: How would this happen where a PendingFileRecoverable is that large in size? Is there any recovery from this point?

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
```diff
@@ -147,71 +191,226 @@ public void flatMap(
...
+        // a different partition than the original attempt did, and - because load jobs are
+        // idempotent on jobId, not on content - the COPY job would then double-count any file the
+        // first attempt had already loaded into a different partition's temp table.
+        final List<PendingFileRecoverable> sortedFiles =
+                files.stream()
+                        .sorted(Comparator.comparing(file -> file.getPath().toUri().toString()))
+                        .collect(Collectors.toList());
+
+        // bin-pack files into partition
+        final List<List<PendingFileRecoverable>> partitions = new ArrayList<>();
```
drossos, line 267: Was going to suggest more optimized size bin-packing, but the benefits are really not there considering majority of pending files will be in similar size so doing this single pass through approach makes most sense from optimization perspective.

Just leaving comment here for posterity incase other reviewer thinks similar.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
```diff
@@ -147,71 +191,226 @@ public void flatMap(
...
+            LOG.info(
+                    "Loading partition {} ({} files) into temp table {}",
+                    partitionIndex,
+                    partitionFiles.size(),
+                    formatTableId(tempTableId));
+
+            final JobConfiguration tempLoadConfig =
+                    applyParquetFormatOptions(LoadJobConfiguration.newBuilder(tempTableId, gcsUris))
+                            .setWriteDisposition(JobInfo.WriteDisposition.WRITE_TRUNCATE)
+                            .setCreateDisposition(JobInfo.CreateDisposition.CREATE_IF_NEEDED)
```
drossos, line 356: I am pretty sure this is already true for the direct-loading as well, but can you confirm the permission for `bigquery.tables.create` is required for direct loading?

I am just making sure that if that permission is new requirement for "large-workloads" then we document it in the configuration.

drossos (commented):
Looks great and makes a lot of sense to me, minor comments / clarifications requested . 👍

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
fqshopify: > "How would this happen where a PendingFileRecoverable is that large in size?"

It can't; `SizeBasedCheckpointRollingPolicy` caps part files at 1.5 GB (well below the 15 TB BQ load-job cap). 
I've removed the check now; it was purely/overly defensive.
We may need to bring it back in the future if/when we expose the rolling-policy size as a config but it can wait till then.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
fqshopify: Agreed. Also another important dimension to remember is the bin-packing strategy needs to be deterministic across retries.

flink-2.1-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/indirect/BigQueryLoadJobOperator.java
fqshopify: Hmmmm this is a good question. 

**Single-partition (direct) load** uses the default `CREATE_IF_NEEDED` disposition, so `bigquery.tables.create` is needed in that case. 

I'm now questioning if this a good idea though. I notice the regular sink has a config option that defaults to false: 

```
    /** [OPTIONAL, Sink Configuration] Boolean flag controlling table creation in the sink. */
    public static final ConfigOption<Boolean> ENABLE_TABLE_CREATION =
            ConfigOptions.key("write.enable-table-creation")
                    .booleanType()
                    .defaultValue(false)
                    .withDescription("Enable table creation in sink");
```

fqshopify: Ok; I've switched both the direct-load path and the multi-partition final copy job to CREATE_NEVER so the load job fails loudly on a missing destination table.
    
This aligns the indirect path with my intentions; BigQuerySinkConfig.validateIndirect already rejects enableTableCreation=true outright so indirect was already contracted to "table must exist." CREATE_NEVER just enforces that contract server-side, and means bigquery.tables.create is not required for the final destination table being written to by indirect writes today. 

Note: The temp-table loads in the multi-partition path keep CREATE_IF_NEEDED since they create the temp tables in the configured tempDataset (so bigquery.tables.create is still required on the temp dataset).

Planned follow-up PR will plumb enableTableCreation through to the indirect path (drop the validateIndirect rejection, flip the disposition based on the flag).

drossos (approved):
LGTM 👍  great stuff. Will leave Google to have their own review as well

yalimu-g (approved):

PR #327, opened 2026-06-09
Fix LIKE filter pushdown never being applied

### LINKED ISSUE #328: bug: `LIKE 'prefix%'` filter pushdown is never applied
# `LIKE 'prefix%'` filter pushdown is never applied (regex can't match the quoted literal)

## Summary
`BigQueryRestriction.convertLike` is meant to push down `LIKE 'prefix%'` ("starts with") filters to
BigQuery as a row restriction, but **no `LIKE` filter is ever pushed down**.

## Root cause
The pattern literal is resolved through `convertLiteral`, which wraps strings in single quotes
(`addSingleQuotes`), so `foo%` becomes `'foo%'`. `convertLike` then matches that quoted value against:

```java
private static final Pattern STARTS_WITH_PATTERN = Pattern.compile("([^%]+)%");
```

with `matcher.matches()`, which requires the entire string to end in `%`. The quoted literal ends in
`'`, not `%`, so `matches()` always returns `false` and `convertLike` returns `Optional.empty()`.
Prefix `LIKE` predicates are therefore silently never pushed down — queries scan/return more rows
from BigQuery than necessary.

There is also a latent second bug: the result was built as `name LIKE '<group1>'`, dropping the
trailing `%` — turning a prefix match into an exact match (a more restrictive predicate).

fqshopify:
Fixes: https://github.com/GoogleCloudDataproc/flink-bigquery-connector/issues/328

### DIFF
--- a/flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/restrictions/BigQueryRestriction.java
+++ b/flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/table/restrictions/BigQueryRestriction.java
@@ -191,18 +191,15 @@ public class BigQueryRestriction {
 
         if (left instanceof FieldReferenceExpression && right instanceof ValueLiteralExpression) {
             String name = ((FieldReferenceExpression) left).getName();
-            return convertLiteral((ValueLiteralExpression) right)
+            return ((ValueLiteralExpression) right)
+                    .getValueAs(String.class)
                     .flatMap(
-                            lit -> {
-                                if (lit instanceof String) {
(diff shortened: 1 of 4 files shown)


fqshopify:
cc: @yalimu-g @BoqianShi

fqshopify:
RE all the "unescaped quotes" comments raised by AI review above: 

This concern applies to all of the code in this file and needs a broader fix beyond the scope of this PR. I will leave this to Google team to address as I'm moving to another project. Issue filed here: https://github.com/GoogleCloudDataproc/flink-bigquery-connector/issues/329

yalimu-g (approved):

yalimu-g:
Please run 'mvn spotless:apply' to fix these violations.

PR #332, opened 2026-06-19
Null-check unsupported function definitions in BigQueryRestriction.convert

drossos:
# Context
BigQueryRestriction.convert looks up the call's FunctionDefinition in the FILTERS map and then does switch (op) without a null check. Any function definition not in the map (e.g. COALESCE, CASE, IF, CAST) unboxes a null Operation and throws NullPointerException during job planning.

I believe the original intention was for this to be caught by the `default` case, but as the switch calls `ordinal()` on null, the NPE is thrown before it ever gets there.

# Fixes

- By adding a null check before the other cases this prevents the hitting of an NPE on unknown operation
- Instead of throwing exception on default, we just follow the correct pattern of returning Optional.empty() which indicates to the caller that this filter needs to be handled on the Flink side. NOTE, this default case is defensive and never hit in practice.


BoqianShi (changes requested):
LGTM, can you fix the cla problem then I can merge it?

drossos:
@BoqianShi Sorry for delay on response here, just updated CLA + comments

BoqianShi (approved):
LGTM, thanks!

drossos:
@BoqianShi huge thanks! Can you please merge this in when convenient for you (I am unable to)?

PR #334, opened 2026-07-10
Add indrect mode to integration test

yalimu-g:


flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -361,6 +364,13 @@ public static void main(String[] args) throws Exception {
                     case "bounded":
                         sourceDatasetName = parameterTool.getRequired("bq-source-dataset");
                         sourceTableName = parameterTool.getRequired("bq-source-table");
+                        String writeModeStr = parameterTool.get("write-mode", "STORAGE_WRITE_API");
+                        String tempGcsPath =
+                                parameterTool.get(
+                                        "temp-gcs-path", parameterTool.get("temporary-gcs-bucket"));
```
davidrabinowitz, line 376: does `temporary-gcs-bucket` always exist?

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -470,6 +490,26 @@ private static void runBoundedFlinkJobWithSink(
                                         : DeliveryGuarantee.AT_LEAST_ONCE)
                         .streamExecutionEnvironment(env);
 
+        boolean isIndirect =
```
davidrabinowitz: why do we need this complex condition? Why not just rely on the writeMode ? This condition may cause user to configure writeMode=INDIRECT but  STORAGE_WRITE_API will be used due toa missing tempGcsPath. If INDIRECT is set but tempGcsPath is missing an error should be thrown

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -361,6 +364,13 @@ public static void main(String[] args) throws Exception {
                     case "bounded":
                         sourceDatasetName = parameterTool.getRequired("bq-source-dataset");
                         sourceTableName = parameterTool.getRequired("bq-source-table");
+                        String writeModeStr = parameterTool.get("write-mode", "STORAGE_WRITE_API");
```
davidrabinowitz: write mode should be an ENUM (DIRECT, INDIRECT)

davidrabinowitz (changes requested):

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
yalimu-g: yes, if temp-gcs-path does not exist(e.g. flink 1.17) it will fallback to temporary-gcs-bucket.

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
yalimu-g: Write mode does not have Direct. But I changed this to boolean for checking if this is indirect or not.

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
yalimu-g: Ack - removed extra check on the tempGcsPath.

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -469,6 +495,20 @@ private static void runBoundedFlinkJobWithSink(
                                         ? DeliveryGuarantee.EXACTLY_ONCE
                                         : DeliveryGuarantee.AT_LEAST_ONCE)
                         .streamExecutionEnvironment(env);
+        if (isIndirect) {
```
BoqianShi, line 498: The indirect sink is batch-only, but this path leaves the environment in Flink 2.1’s default `STREAMING` mode. `BigQueryIndirectSink.addPostCommitTopology()` will therefore throw `INDIRECT write mode is only supported in BATCH execution mode` when the topology is built. Could we call `env.setRuntimeMode(RuntimeExecutionMode.BATCH)` when `isIndirect` is true?

flink-2.1-connector-bigquery/flink-connector-bigquery-integration-test/src/main/java/com/google/cloud/flink/bigquery/integration/BigQueryIntegrationTest.java
```diff
@@ -469,6 +495,20 @@ private static void runBoundedFlinkJobWithSink(
...
+            sinkConfigBuilder
+                    .schemaProvider(destSchemaProvider)
+                    .writeMode(WriteMode.INDIRECT)
+                    .tempGcsPath(tempGcsPath)
+                    .tempProject(tempProject != null ? tempProject : destGcpProjectName)
+                    .tempDataset(tempDataset != null ? tempDataset : destDatasetName)
+                    .jobProject(jobProject != null ? jobProject : destGcpProjectName)
+                    .bulkWriterFactory(
+                            AvroWriters.forGenericRecord(destSchemaProvider.getAvroSchema()))
+                    .formatOptions(FormatOptions.avro());
```
BoqianShi, line 510: Please enable Avro logical-type conversion here. The destination schema represents `ts TIMESTAMP` as an Avro `long` annotated with `timestamp-micros`, but [BigQuery ignores most Avro logical types by default](https://docs.cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro#logical_types). Plain `FormatOptions.avro()` therefore interprets `ts` as `INTEGER`, which can make the load into the pre-created `TIMESTAMP` column fail. Could we use `AvroOptions.newBuilder().setUseAvroLogicalTypes(true).build()`?

PR #337, opened 2026-07-27
Add timeout handling for buffered and default writer

yalimu-g:
When incoming data traffic pauses or is low-volume for >10 minutes, BigQuery's Storage Write API server closes the inactive gRPC stream. When new records arrive or Flink triggers a checkpoint, the connector attempts to validate pending write futures using an unbounded future.get() call. Because the underlying gRPC stream is dead or hung, future.get() blocks indefinitely until Flink's execution.checkpointing.timeout triggers an InterruptedException, failing the task.

This PR re-initialize StreamWriter on Connection Failure / Timeout: Catch TimeoutException and connection closed / gRPC UNAVAILABLE exceptions, close the stale streamWriter, and recreate a fresh streamWriter (createStreamWriter(...)).

### DIFF
--- a/flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
+++ b/flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
@@ -84,7 +84,7 @@ abstract class BaseWriter<IN> implements SinkWriter<IN> {
     // Multiply 0.95 to keep a buffer from exceeding payload limits.
     private static final long MAX_APPEND_REQUEST_BYTES =
             (long) (StreamWriter.getApiMaxRequestBytes() * 0.95);
-    protected static final long DEFAULT_APPEND_RESPONSE_TIMEOUT_SECONDS = 300L;
+    protected static final long DEFAULT_APPEND_RESPONSE_TIMEOUT_SECONDS = 420L;
 
     // Number of bytes to be sent in the next append request.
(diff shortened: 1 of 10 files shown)


flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
```diff
@@ -84,6 +84,7 @@ abstract class BaseWriter<IN> implements SinkWriter<IN> {
     // Multiply 0.95 to keep a buffer from exceeding payload limits.
     private static final long MAX_APPEND_REQUEST_BYTES =
             (long) (StreamWriter.getApiMaxRequestBytes() * 0.95);
+    protected static final long DEFAULT_APPEND_RESPONSE_TIMEOUT_SECONDS = 300L;
```
BoqianShi: 300s matches `ConnectionWorker.MAXIMUM_REQUEST_CALLBACK_WAIT_TIME` exactly, which means the connector timeout and the client library's internal cleanup could race — consider bumping to ~360-420s to let the library finish its recovery first.

BoqianShi (approved):
LGTM with NIT comments

flink-1.17-connector-bigquery/flink-connector-bigquery/src/main/java/com/google/cloud/flink/bigquery/sink/writer/BaseWriter.java
yalimu-g: thanks. Done

PR #353, opened 2026-09-16
Fix view materialization job project

fqaiser94:
## Problem

When reading a view with a custom `read.views.billing-project`, the materialization job is submitted under the BigQuery client's default project instead of the configured billing project. If the client's credentials lack `bigquery.jobs.create` in that project, the source's SplitEnumerator fails with a permission error and restarts indefinitely; if they don't, the job silently runs (and bills) in the wrong project.

Root cause: `materializeView` rebuilds the client with `setQuotaProjectId(billingProject)`, which only sets the `x-goog-user-project` header for API request quota attribution. `BigQuery#create(JobInfo)` places jobs in the client's default project when `JobInfo` has no explicit `JobId`, so the option never affected job placement or billing.

## Fix

When `read.views.billing-project` is set, attach an explicit `JobId` carrying the billing project to the materialization job. Everything else is unchanged:

- when the option is unset, behavior is identical to released versions: the job runs under the client's default project 
- the quota-project client (`setQuotaProjectId`) is preserved as-is;

## Behavior change

Only for users who set `read.views.billing-project`: materialization jobs move from the client's default project to the configured billing project i.e. the option now does what it is documented to do. Users who have not configured the option see no change.

## Testing

- Unit  tests
- Manual E2E against real BigQuery views across two projects with `read.views.billing-project` set: the job runs and is listed in the billing project, nothing is submitted to the client's default project, and query results are correct.


BoqianShi:
Could we preserve the existing default when `read.views.billing-project` is unset?

Honoring an explicitly configured billing project fixes the reported issue, but falling back to the materialization/source project also changes behavior for users who haven’t configured this option. The project chosen for temporary-table storage doesn’t necessarily represent the intended query-execution project. Existing cross-project deployments could start failing on `bigquery.jobs.create`, or have their billing and reservation usage change without any configuration changes.

I’d suggest using the explicit billing project when provided and otherwise retaining the BigQuery client’s default project. If we want to change that default, could we treat it as a separate behavior change with migration guidance?

BoqianShi (changes requested):
Please preserve the existing default when read.views.billing-project is unset.

fqaiser94:
> Could we preserve the existing default when read.views.billing-project is unset?

Done. 


> Honoring an explicitly configured billing project fixes the reported issue, but falling back to the materialization/source project also changes behavior for users who haven’t configured this option. 

Sorry, I didn't realize the views functionality had already been released. Agree that making a default behaviour change now would require a major version release and it isn't necessary.

BoqianShi:
Thanks for the edit!

One small documentation cleanup: the PR description still mentions materialize_bqc_* job IDs, but the updated code generates random IDs.

fqaiser94:
> One small documentation cleanup: the PR description still mentions materialize_bqc_* job IDs, but the updated code generates random IDs.

PR description updated now.

BoqianShi (approved):
LGTM, thank you!

